PR #23, opened 2021-03-17
style: improved styling to match Swarm branding

wendydv1989:


src/components/NavBar.tsx
```diff
@@ -40,21 +39,21 @@ export default function SideBar(props: any) {
 
   return (
     <div>
-      <AppBar position="fixed" className={classes.appBar}>
-        <Toolbar style={{display: 'flex'}}>
+      <div style={{ display: 'fixed' }} className={classes.appBar}>
```
vojtechsimetka, line 42: why is this changed to div? Also don't use `style` here please, instead add any styling into `appBar` object.

vojtechsimetka (approved):

wendydv1989:
with love

src/components/SideBar.tsx
```diff
@@ -2,46 +2,47 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 
 import { createStyles, Theme, makeStyles } from '@material-ui/core/styles';
-import { ListItemText, ListItemIcon, ListItem, Divider, List, Drawer, Link as MUILink} from '@material-ui/core';
+import { ListItemText, ListItemIcon, ListItem, Divider, List, Drawer, Link as MUILink, ListItemSecondaryAction } from '@material-ui/core';
 import { OpenInNewSharp } from '@material-ui/icons';
-import { Activity, FileText, DollarSign, Share2, Settings } from 'react-feather';
+import { Activity, FileText, DollarSign, Share2, Settings, AlignJustify } from 'react-feather';
```
matmertz25, line 7: It looks like `AlignJustify` might be an unused import

src/components/NavBar.tsx
matmertz25: I think if we are going to use a `div` tag here in stead of the AppBar component we need to remove the margin from the top of the `Dashboard` component content otherwise everything gets offset for me. 

https://github.com/ethersphere/bee-status/blob/34d2dfda5a89fc0b103d6aabbd2c1af5e06215a6/src/layout/Dashboard.tsx#L16

image

matmertz25 (approved):
Looks good, thanks for updating to the brand styling

centerorbit:
Merging because these new styles/icons are merged into electron builds.

PR #24, opened 2021-03-21
feat: Adding electron builds for Mac, Win, and Linux

centerorbit:
BREAKING CHANGE: 'start' and 'electron-dev' now use port 3000 (was non-standard 3001)
fix: Loading of assets in non-root serves/hosting (header sets <base>)
fix: Redirect to app relative 'home' if loading app from "index.html" (catch-all in router)

Windows exe, MacOS dmg, Linux (deb,rpm). AppImage and Snap are easy additionas as well.

The routing was changed to have a 'catch-all' to route to '/' due to how
initial page loads work with electron.

A bash script was added to use docker for building windows exe on linux.
This is because otherwise you need Wine installed on your system, which
is a very large dependency.


centerorbit:
@matmertz25 I changed the default port back to 3000, as that's most common/standard for React.

centerorbit:
I had rebased in @wendydv1989 changes into my branch, so these are showing up as changes as well but in reality it's the same as what's on master.

public/electron.js
```diff
@@ -0,0 +1,75 @@
...
+// Some APIs can only be used after this event occurs.
+app.whenReady().then(() => {
+  createWindow();
+
+  if (isDev) {
+    installExtension(REACT_DEVELOPER_TOOLS)
+      .then(name => console.log(`Added Extension:  ${name}`))
+      .catch(error => console.log(`An error occurred: , ${error}`));
+  }
+});
```
nugaon, line 55: it is fine (though see my comment regarding to `installExtension` part), but within our repositories we use some conventions in order to the implementation seem unified.
it should be moved into an async function where please use `await` and `try ... catch`.

public/electron.js
```diff
@@ -0,0 +1,75 @@
...
+
+
+// Conditionally include the dev tools installer to load React Dev Tools
+let installExtension, REACT_DEVELOPER_TOOLS;
+
+if (isDev) {
+  const devTools = require("electron-devtools-installer");
+  installExtension = devTools.default;
+  REACT_DEVELOPER_TOOLS = devTools.REACT_DEVELOPER_TOOLS;
+}
```
nugaon, line 14: This code snippet is strange considering the usage later
https://github.com/ethersphere/bee-status/blob/806c48174876aa32e392cfbc467576ea532ee93e/public/electron.js#L50-L54
it seems this above snippet is the only part where you use `installExtension` and `REACT_DEVELOPER_TOOLS`.
so defining these two variables in the outer scope is unjustified and split this workflow pollutes the code (and check this condition two times is unnecessary on this same level) 
these two part must be merged

win-build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
nugaon, line 1: this is a util script, should be moved to some subdirectory

nugaon (changes requested):
Usually the method definitions precede the execution part of the file, it also caused some little problems.
But the real problem for me is I thought the electron process also would be written in TypeScript since our other repositories (like this one as well before) have been created by in that.

package.json
```diff
@@ -51,15 +55,43 @@
...
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "electron-dev": "concurrently \"BROWSER=none npm run start\" \"wait-on http://localhost:3000 && electron .\"",
+    "electron-build": "react-scripts build && electron-builder build",
+    "electron-build:linux": "react-scripts build && electron-builder build -l --publish never",
+    "electron-build:linux:deb": "react-scripts build && electron-builder build --linux deb --publish never",
+    "electron-build:linux:rpm": "react-scripts build && electron-builder build --linux rpm --publish never",
+    "electron-build:win": "react-scripts build && electron-builder build -w --publish never",
+    "electron-build:mac": "react-scripts build && electron-builder build -m dmg --publish never"
```
nugaon, line 78: instead of `react-scripts build` call `build` defined script

centerorbit:
> Usually the method definitions precede the execution part of the file, it also caused some little problems.
> But the real problem for me is I thought the electron process also would be written in TypeScript since our other repositories (like this one as well before) have been created in that.

@nugaon I agree with the sentiment of wanting it all in TypeScript, except the `electron.js` file runs on the node.js side of electron (vs the browser side) which is used just to bootstrap the window (which is then all React). The trouble is that the app was created using [Create React App](https://create-react-app.dev/), and Electron was layered in afterwards, so all Webpack and Babel config is manged by this utility/package. The only way to reasonably change this config (to add a typescript build process for the electron.js file)  is to 'eject' the app, which then makes updating the app in the future much more difficult. So I've been trying to not eject the app unless absolutely necessary.

With all that considered, it is more pain than gained to maintain the one JS file to bootstrap the windowing rather than eject-and-alter the config _or_ install a second set of webpack just to transpile the one file.

public/electron.js
```diff
@@ -0,0 +1,75 @@
...
+
+if (require("electron-squirrel-startup")) {
+  app.quit();
+}
+
+function createWindow() {
+  // Create the browser window.
+  const win = new BrowserWindow({
```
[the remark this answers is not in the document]
centerorbit: The package.json defines the icons in the "build" section of the config. You'll see definitions for each OS target.

src/routes/routes.tsx
```diff
@@ -21,7 +21,9 @@ const BaseRouter = (props: any) => (
         <AppRoute exact path='/peers/' layout={ Dashboard } component={Peers}/>
         <AppRoute exact path='/accounting/' layout={ Dashboard } component={Accounting}/>
         <AppRoute exact path='/settings/' layout={ Dashboard } component={Settings}/>
+        {/* A catch-all because electron build initial load is: /absolute/path/index.html */}
+        <Route path='*'><Redirect to='/'/></Route>
```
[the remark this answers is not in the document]
centerorbit: If the goal is to have a 'desktop' like application experience, a "not found" doesn't make much sense (since it's not a client-server relationship). There are no browser-back buttons in the electron build either, so if a user were to hit a 404 page they may get stuck to the point of having to exit the app.

There was another reason I added this, though, but the <base> change I made below may have solved the issue as well. Let me investigate more.

vojtechsimetka:
>With all that considered, it is more pain than gained to maintain the one JS file to bootstrap the windowing rather than eject-and-alter the config or install a second set of webpack just to transpile the one file.

Thanks @centerorbit for the PR and the provided reasoning. The reason this was started as Create React App is because we want to package it directly into Bee node (in addition to the electron app). As you said I also don't want to eject if possible and hence from my end I could live with some JS files.

Would it be possible to have a two package.json structure? So that the website build is not polluted with the electron and vice versa. This would be great separation and might solve some of the identified issues https://www.electron.build/tutorials/two-package-structure

centerorbit:
> > With all that considered, it is more pain than gained to maintain the one JS file to bootstrap the windowing rather than eject-and-alter the config or install a second set of webpack just to transpile the one file.
> 
> Thanks @centerorbit for the PR and the provided reasoning. The reason this was started as Create React App is because we want to package it directly into Bee node (in addition to the electron app). As you said I also don't want to eject if possible and hence from my end I could live with some JS files.
> 
> Would it be possible to have a two package.json structure? So that the website build is not polluted with the electron and vice versa. This would be great separation and might solve some of the identified issues https://www.electron.build/tutorials/two-package-structure

Yeah! I'll look into that.

centerorbit:
@nugaon @vojtechsimetka I believe I've addressed all of the review notes accept for the 2 package.json files. I'll be looking into that next.

centerorbit:
> @nugaon @vojtechsimetka I believe I've addressed all of the review notes accept for the 2 package.json files. I'll be looking into that next.

@vojtechsimetka I've been reading up on the purpose of the two package.json approach and it's not really designed for the purpose that you're proposing. It was designed back around version 5 of electron-builder for native C dependencies to be compiled for specific architecture targets. By version 8 this was mostly made obsolete. We're using version 22. I worry that if we choose to use this technique for its unintended purpose, that it may be deprecated in the near future because it's actual purpose is no longer needed.

NPM has a concept of 'optionalDependencies', which we may be able to leverage, but not only would a user need to specify a `--no-optional` flag while installing, I'd also worry that this would wreak havoc on the CI because in those cases the 'optionals' are no longer optional... which could lead to odd behavior.

Another option (one that I'm most favorable towards), is similar to the two package.json would be to write a few scripts that would run a build, and then copy the built assets, along with other needed files into a sub-directory and then perform the electron build there. It wouldn't be using the electron-builder 'mechanism' of two package.jsons, but it'd essentially be accomplishing the same thing.

Part of me feels like this is over engineering a problem that doesn't really need to be solved. While the near-term use is people will need to install and compile this application themselves to run it; the actual goal is to distribute the Electron app itself. If just having the website build running, with no electron, we could produce a single HTML file that is all of the assets bundled into it. The user would only need to download that one file and double-click it to open in a browser, or we could package the pre-compiled assets into a simple docker container.

src/routes/routes.tsx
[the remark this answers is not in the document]
centerorbit: @nugaon I looked into the other reason I added this, and it's because when electron first loads, the path it starts is an absolute path to the location of the application, and then its asar file (a bundle of web assets) and _then_ the location of the index.html file that's built from React.

If the electron had a browser bar, it's location would be something like : `file://home/andy/.config/bee-status/package/app.asar/build/index.html`

Which obviously doesn't match the `exact path='/'` of the first route. I add the catch-all to redirect to the `/` so that the home dashboard loads properly.

Before this addition, the app would start and it'd just be stuck on a blank screen because the route matched nothing, so nothing loaded.

centerorbit:
@nugaon I had a call with @vojtechsimetka today, and he had an excellent suggestion for how to make the electron.js file a ts file without having to eject the app. You'll see in the `package.json` I added a script to invoke `tsc` directly to compile the electron.ts file (which is now located in the 'src' directory).

Take a look and let me know if there are any additional changes that you'd like.

package.json
```diff
@@ -44,22 +48,52 @@
...
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "electron-dev": "npm run electron-build-bootstrap && concurrently \"BROWSER=none npm run start\" \"wait-on http://localhost:3000 && electron .\"",
+    "electron-build": "npm run build && electron-builder build",
+    "electron-build:linux": "npm run build && electron-builder build -l --publish never",
+    "electron-build:linux:deb": "npm run build && electron-builder build --linux deb --publish never",
+    "electron-build:linux:rpm": "npm run build && electron-builder build --linux rpm --publish never",
+    "electron-build:win": "npm run build && electron-builder build -w --publish never",
+    "electron-build:mac": "npm run build && electron-builder build -m dmg --publish never",
+    "electron-build-bootstrap": "tsc --esModuleInterop --outDir public src/electron.ts"
```
vojtechsimetka, line 80: Rather than changing the `build` which I would love to keep just for the react website building, lets change the `electron-bootstrap` to run the `npm run build`. So it would look as follows:

```json
"scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject",
    "electron-bootstrap": "tsc --esModuleInterop --outDir public src/electron.ts",
    "electron-build": "electron-bootstrap && npm run build && electron-builder build",
    "electron-build:linux": "electron-build -- -l --publish never",
    "electron-build:linux:deb": "electron-build -- --linux deb --publish never",
    "electron-build:linux:rpm": "electron-build -- --linux rpm --publish never",
    "electron-build:win": "electron-build -- -w --publish never",
    "electron-build:mac": "electron-build -- -m dmg --publish never",
    "electron-dev": "npm run electron-bootstrap && concurrently \"BROWSER=none npm run start\" \"wait-on http://localhost:3000 && electron .\""
  },
```

nugaon:
> install a second set of webpack just to transpile the one file.

separated webpack config would be better to compile electron.ts, that's what I was thinking about first as we wanted to use this setup in our subsequent electron projects, but I accept this project does not need this feature (at the current phase) so it's fine to keep it like now.
Nevertheless there are some other issues
- remove public/electron.js
- icons don't show up on the taskbar on Linux.
- if I run the AppImage and I click on a URL, the whole window goes blank.
regarding to menubar:
- view -> reload/force-reload makes the whole app blank, you cannot restore.
- window zoom does not do anything
- help section is empty

matmertz25:
@vojtechsimetka is there any interest in establishing the electron / desktop application in its own repository? Ideally this repo can purely build a webapp from which the static files can be used in the electron application. 

My main reasoning for this is as the "status" application matures the code bases will continue to diverge and a good deal of the functionality that we would want to build into the desktop version like using the file system apis to pull node configuration files, edit config files, execute commands, potentially bundling a bee node, etc, will be very different from the functionality that the web application portion seeks to deliver. It also keeps a clean separation of dependencies and gives the desktop version room to grow rather than just be pretty much a pure web app wrapped in a desktop window. 

@centerorbit  thoughts?

centerorbit:
@matmertz25, @vojtechsimetka and I had talked about this possibility a bit a week ago. The trouble is that a lot of duplicate code might be needed to maintain each unless a third repo is established that contains a bunch of the components used that both projects would then leverage, which would add a headache to the maintenance of both repos.

From my understanding, they are no longer going to embed a version of the status app inside a bee-node itself, so the only place this code will run is in an electron build itself... unless someone sets up their own local dev/web build of the SPA. Which then begs the question, why not just make this the electron+react project?

centerorbit:
@nugaon below are my comments, I'll make changes for the "electron.js item", but I also think project/repo direction needs to be defined a bit more.

When I undertook this effort to wrapper this alpha-state project in electron my initial understanding was the goal was to get a 'proof-of-concept' electron build working that would later be improved upon as the project matured. This PR and feedback is now showing that the project/repo goal needs to be better defined, and that the requirements for the electron build need more definition.

It feels like the further I get in the production of this build, the more fine-grained the feedback is becoming, which is no longer a proof-of-concept and more like a full-featured deliverable. This would be fine if the whole scope was initially defined, but right now it does not feel like iterative development.

IF this repos purpose is to become a "bee-status electron app" exclusively (no web build), then this PR should be merged sooner and 'issues' be logged against features that are desired to change.

If that is not the case, and web view is to be supported or electron app with a separate repo, or a shared component library, that all needs larger discussion.

> * remove public/electron.js
I thought I had removed and gitignored this, it must have slipped back in, I'll fix it.

> * icons don't show up on the taskbar on Linux.
What Linux distro and build-type are you using? Using the Deb package with a Debian based distro works on my machine.

> * if I run the AppImage and I click on a URL, the whole window goes blank.
This may be one of the first instances that @matmertz25 was bringing up about "native" APIs. When you click on a link in Electron you want a web browser to open up. When you click on a link in a browser you want a new tab. To make this link behave properly we may have to either change how the URL is structured or add a special handler which means this repo would likely 'become' an electron-app only repo.

Also, I have not explicitly tested/added AppImage builds, but I can look into this. I only got Windows-exe, Mac-dmg, Linux-rpm, and Linux-deb working.

What build targets were you thinking about supporting with this app? The options are:

Linux:
AppImage, snap, deb, rpm, freebsd, pacman, p5p, apk, 7z, zip, tar.xz, tar.lz, tar.gz, tar.bz2, dir

Windows:
nsis, nsis-web (Web installer), portable (portable app without installation), appx, msi, squirrel, 7z, zip, tar.xz, tar.lz, tar.gz, tar.bz2, dir

macOS:
dmg, mas, mas-dev, pkg, 7z, zip, tar.xz, tar.lz, tar.gz, tar.bz2, dir


>   regarding to menubar:
I'm not sure what direction you want to take on the menubar. What is in there are the defaults that Electron provides, which are not really specific to the app. To 'fix' these issues I can simply remove all menubar options, and then PRs can be brought back at a later date to add in functionality that is actually desired.

> * view -> reload/force-reload makes the whole app blank, you cannot restore.
> * window zoom does not do anything
> * help section is empty

wendydv1989:
Awesome feedback. 
I can do the electron part if you feel like focussing on the react part.

nugaon:
> @nugaon below are my comments, I'll make changes for the "electron.js item", but I also think project/repo direction needs to be defined a bit more.

> When I undertook this effort to wrapper this alpha-state project in electron my initial understanding was the goal was to get a 'proof-of-concept' electron build working that would later be improved upon as the project matured. This PR and feedback is now showing that the project/repo goal needs to be better defined, and that the requirements for the electron build need more definition.

> It feels like the further I get in the production of this build, the more fine-grained the feedback is becoming, which is no longer a proof-of-concept and more like a full-featured deliverable. This would be fine if the whole scope was initially defined, but right now it does not feel like iterative development.

> IF this repos purpose is to become a "bee-status electron app" exclusively (no web build), then this PR should be merged sooner and 'issues' be logged against features that are desired to change.

I had two comments on your PR:
1. fixes and corrections of your code + I said it should be in typescript as the other part of the codebase (not really much overhead)
2. I pointed on bugs that does not work in the electron app and should be fixed because some parts are not functional with this setup. Additionally I said it could be nicer if the previously mentioned ts compilation would happen by webpack, but I said it does not matter and totally fine as it is.

IMO a setup which bootstraps something also should use those configurations which work correctly within that environment that it creates.

> Also, I have not explicitly tested/added AppImage builds, but I can look into this. I only got Windows-exe, Mac-dmg, Linux-rpm, and Linux-deb working.

I was not there at the initial conversation about the project, but I don't think it is justified to point to the other they don't specify the targets. It generates the AppImage to the dist folder by your configuration which does not work. If you say you cannot test every image type that electron produces by default it is fine, tell this info to that person who negotiated with you this task and discuss whether even it is needed or not. I think this is part of the deal and teamwork from both sides.

required build targets would be:

Linux:
AppImage, deb

Windows:
portable (portable app without installation)

macOS:
dmg, 

> I'm not sure what direction you want to take on the menubar. What is in there are the defaults that Electron provides, which are not really specific to the app. To 'fix' these issues I can simply remove all menubar options, and then PRs can be brought back at a later date to add in functionality that is actually desired.

it's fine, I think we can remove the menubar.

nugaon:
@centerorbit for the icon problem try out defining the icon in the `BrowserWindow` init: https://github.com/ethersphere/bee-status/pull/24#discussion_r598780563

vojtechsimetka:
Hi @centerorbit I think @nugaon already clarified what needs to be done (we talked about it briefly). So just to summarise:

The goal of this PR is to wrap the app into electron and make sure it builds for Linux, macOS and Windows (+ manually test the outputs, we can help with this :) )

> https://github.com/ethersphere/bee-status/pull/24#issuecomment-811297537 by @centerorbit 
> When I undertook this effort to wrapper this alpha-state project in electron my initial understanding was the goal was to get a 'proof-of-concept' electron build working that would later be improved upon as the project matured. This PR and feedback is now showing that the project/repo goal needs to be better defined, and that the requirements for the electron build need more definition.

Apologies this is my fault. At the beginning it indeed was a proof-of-concept as we wanted to meet a specific deadline. I did not communicate with you that it is no longer the case so the expectation is that this PR is solid.

Here are the milestone which hopefully will make this clear:
- Webpage MVP [0.1.0](https://github.com/ethersphere/bee-status/milestone/1)
- Electron app [0.2.0](https://github.com/ethersphere/bee-status/milestone/2)
- Bee Desktop [0.3.0](https://github.com/ethersphere/bee-status/milestone/3)

> https://github.com/ethersphere/bee-status/pull/24#issuecomment-811161750 by @centerorbit 
> From my understanding, they are no longer going to embed a version of the status app inside a bee-node itself, so the only place this code will run is in an electron build itself... unless someone sets up their own local dev/web build of the SPA. Which then begs the question, why not just make this the electron+react project?
>
> So, I guess my thought is if the end-goal is an Electron app, keep them one in the same to simplify the ecosystem until a real need to have web app that can monitor nodes is needed. Which, in that case would likely target multiple nodes as a dashboard to a cluster, and would require a much different design than what we're currently headed towards.

Yes this will not be embedded in bee so yes the end-goal is Electron app. However, IMHO we are way closer to make this useful as a standalone website first so lets finish that target. That way people will be able to already use it while make this purely Electron app (when the features we wont can only be achieved by going pure electron).

Cafe137:
Continuation in https://github.com/ethersphere/bee-dashboard/pull/171

PR #30, opened 2021-03-30
feat: UI improvements and fixes, organize status checks

matmertz25:
Adds the following features / functionality

- Table pagination for accounting data tables
- split theme configuration into separate file
- add swarm gateway as option for file operations 
- split bee debug api and node api status checks
- create custom tab component for reusability
- add loading progress and eth TX link for cashouts


src/pages/accounting/BalancesTable.tsx
```diff
@@ -44,7 +50,7 @@ function BalancesTable(props: IProps) {
                     </TableRow>
                     </TableHead>
                     <TableBody>
-                    {props.peerBalances?.balances.map((peerBalance: PeerBalance, idx: number) => (
+                    {props.peerBalances?.balances.sort((a,b) => b.balance - a.balance).slice(page * 20, page * 20 + 20).map((peerBalance: PeerBalance, idx: number) => (
```
vojtechsimetka, line 53: I don't think this needs to be paginated. It's not usually more than 100 records, we can do a virtualised tables if performance is the motivation. Either way, if you do want to put this as a PR, the `sort` and `splice` should not be done with each re-render. You can use the `useEffect` which would listen on the `props.peerBalances.balances`.

vojtechsimetka:
closing in favour of #37 and #36

PR #35, opened 2021-03-31
style: add eslint configuration as per bee-js

vojtechsimetka:
This adds eslint + prettier configuration that we use in bee-js. Let's discuss if we need to make any changes.  Steps:

- [x] Agree on the rules
- [x] Run the `npm run lint -- --fix`
- [x] Fix all errors that need to be fixed manually
- [x] Fix all warning that need to be fixed manually


.eslintrc
```diff
@@ -0,0 +1,89 @@
+{
+  "extends": ["plugin:@typescript-eslint/recommended", "prettier", "plugin:prettier/recommended"],
```
nugaon, line 2: Add installed react plugin

```suggestion
  "extends": ["plugin:@typescript-eslint/recommended", "prettier", "plugin:prettier/recommended", "plugin:react/recommended"],
```

.eslintrc
```diff
@@ -0,0 +1,89 @@
...
+    "@typescript-eslint/no-unused-expressions": "error",
+    "@typescript-eslint/member-delimiter-style": [
+      "error",
+      {
+        "multiline": {
+          "delimiter": "none",
+          "requireLast": true
+        },
+        "singleline": {
+          "delimiter": "semi",
```
nugaon, line 75: I had a conflict with the prettier about this config so I changed to the following and since then I don't have that problem.

```suggestion
          "delimiter": "comma",
```

.eslintrc
```diff
@@ -0,0 +1,89 @@
...
+    "jest": true,
+    "node": true
+  },
+  "parserOptions": {
+    "sourceType": "module",
+    "ecmaVersion": 2018
+  },
+  "plugins": ["jest"],
+  "rules": {
+    "array-bracket-newline": ["error", { "multiline": true }],
```
nugaon, line 16: I had a conflict with prettier with this setup earlier in one of my project. `"multiline": true` is default so I omitted.

```suggestion
    "array-bracket-newline": ["error", "consistent"],
```

.prettierrc
```diff
@@ -0,0 +1,13 @@
...
+  "useTabs": false,
+  "bracketSpacing": true,
+  "semi": false,
+  "singleQuote": true,
+  "quoteProps": "as-needed",
+  "trailingComma": "all",
+  "endOfLine": "lf",
+  "arrowParens": "avoid",
+  "proseWrap": "always"
+}
```
agazso, line 13: Missing newline at the end ;)

agazso (approved):
Looks good although I did not test it because there were too much reported linter issues

.prettierrc
vojtechsimetka: Thanks, wil add

nugaon (changes requested):
That I noticed `curly: 'error',` rule also defined in some of our projects, but I don't know whether it is neccessary or not.
Nevertheless I would append here.

vojtechsimetka:
> That I noticed `curly: 'error',` rule also defined in some of our projects, but I don't know whether it is neccessary or not.
> Nevertheless I would append here.

There is `"curly": ["error", "multi-line"],`  https://github.com/ethersphere/bee-status/pull/35/files#diff-e7e195d3d8fa5b82fa51ed952f3a179d429cbc46c0432ce1f7b357c54c162822R62 . There is actually configuration error in `bee-js` where we first set it to error and then we allow multiline (which overwrites it) see:
- https://github.com/ethersphere/bee-js/blob/master/.eslintrc.js#L18
- https://github.com/ethersphere/bee-js/blob/master/.eslintrc.js#L63
Setting it as `curly: error` in bee-js would mean we need to change it in a bunch of places. I'd like to keep the `multi-line` option in both.

.eslintrc
vojtechsimetka: This actually clashes on `bee-js` with prettier :(

vojtechsimetka:
@nugaon I've tried to put those two rules you suggested in bee-js but it causes clashes between eslint and prettier. Could we keep it this way and once we make modification to the inter rules on bee-js I'll propagate them here and to other project? :)

 https://github.com/ethersphere/bee-js/pull/245

.eslintrc
nugaon: don't merge then pls

nugaon (approved):

.eslintrc
vojtechsimetka: Yeah don't worry will not merge in `bee-js`. I'll run the lint here and fix the errors though now that we have clarity on the setup :)

vojtechsimetka:
Not sure this can be easily reviewed, I tried to put it into distinct commits. All linter rules are as we agreed on except:

- `"react/react-in-jsx-scope": "off"` React v17+ does not need React in scope for JSX elements
- removed `'no-warning-comments': 'warn'` as I used the `// FIXME:` and `//TODO:` comments to indicate refactors we really need to do. React build sadly treats all warnings as errors when building in CI so I disabled that rule for now.
- `eslint-disable no-nested-ternary` is added in some code blocks and is always prefixed with `// FIXME:` comment as refactoring such code blocks seemed beyond scope of this PR

Please re-review @agazso , probably easiest is to test the app (I already did and everything that works in past still works)

src/components/CodeBlockTabs.tsx
```diff
@@ -1,158 +1,147 @@
...
+    setValue(newValue)
+  }
+
+  useEffect(() => {
+    const os = getOS()
+
+    if (os === 'windows') {
+      setValue(0)
+    } else if (os === 'linux') {
+      setValue(0)
```
agazso, line 58: I don't really understand what's going on here. Maybe using a variable name more descriptive than `value` would be helpful. Also maybe using string literals instead of using values `0` and `1` would convey better what is the intent.

src/components/WithdrawlModal.tsx
```diff
@@ -1,77 +1,70 @@
...
   const handleWithdraw = () => {
     if (amount > 0) {
-        beeDebugApi.chequebook.withdraw(amount)
+      beeDebugApi.chequebook
+        .withdraw(amount)
         .then(res => {
-            setOpen(false);
-            handleToast(`Successful withdrawl. Transaction ${res.transactionHash}`)
+          setOpen(false)
+          handleToast(`Successful withdrawl. Transaction ${res.transactionHash}`)
```
agazso, line 33: I am not a native english speaker and to me `withdrawl` seems a bit strange. I am familiar with `withdrawal` and it seems that is the more commonly used form.

src/pages/accounting/index.tsx
```diff
@@ -1,166 +1,168 @@
...
-    
-    return (
-        <div>
-          {props.nodeHealth?.status === 'ok' && props.health ?
+  return (
+    <div>
+      {
+        // FIXME: this should be broken up
+        /* eslint-disable no-nested-ternary */
+        nodeHealth?.status === 'ok' && health ? (
```
agazso, line 127: I find it's better to extract logic in a variable (e.g. `isNodeHealthy = ...`) and in the JSX only rely on such variables instead of mixing rendering with logic.

src/pages/accounting/index.tsx
```diff
@@ -1,166 +1,168 @@
...
-            <Container style={{textAlign:'center', padding:'50px'}}>
-                <CircularProgress />
-            </Container>
-            :
-            <TroubleshootConnectionCard />
-            }
-        </div>
-    )
+          </div>
+        ) : isLoadingHealth || isLoadingNodeHealth ? (
```
agazso, line 158: Same as above regarding logic in JSX.

src/pages/files/index.tsx
```diff
@@ -30,128 +31,138 @@ const useStyles = makeStyles((theme: Theme) =>
...
-          setUploadingFile(false)
-        })
-    }
+        const downloadUrl = window.URL.createObjectURL(new Blob([res.data]))
+        const link = document.createElement('a')
+        link.href = downloadUrl
+        link.setAttribute('download', 'file.zip') //any other extension
+        document.body.appendChild(link)
+        link.click()
+        link.remove()
```
agazso, line 63: Seems like something to be extracted to a separate function.

src/pages/files/index.tsx
```diff
@@ -30,128 +31,138 @@ const useStyles = makeStyles((theme: Theme) =>
...
-              </div>
-              {inputMode === 'browse' ? 
-              <Paper component="form" className={classes.root}> 
-                  <InputBase
+  return (
+    <div>
+      {
+        // FIXME: this should be broken up
+        /* eslint-disable no-nested-ternary */
+        nodeHealth?.status === 'ok' && health ? (
```
agazso, line 100: Seems like a duplicate of a previous comment. Would be interesting to investigate if the duplication is worth eliminating.

agazso (approved):
Looks good.

I left some comments but they are not necessarily related to this PR. Feel free to ignore them or open issues.

agazso:
I found one issue when testing and it may be unrelated to this PR. When checking the status and no node is connected there is a text displaying `html_url` which seems like a bug.

Screen Shot 2021-04-03 at 13 29 23

PR #38, opened 2021-04-01
refactor: use bee-js for health endpoint

vojtechsimetka:
Replaces the axios call for the Bee endpoint, fixes the interfaces for file upload and download.

The file download does not work well for me, but this PR does not really change it so I'll just create an issue about it as I think that part should anyways be refactored into more react way - see #39


agazso (approved):
I also tried the file upload and it works 😄 
However the browse button doesn't do anything, maybe it's better to remove

vojtechsimetka:
Yeah it's the download that does not work :)

PR #49, opened 2021-04-02
refactor: add theme provider to toggle theme without app reload

### LINKED ISSUE #48: Theme change should not require page reload
Right now changing theme from light to dark requires the page to reload. This is not necessary. The value should be stored in a Theme provider which also exposes a switch between `light` and `dark` theme and is responsible for writing it into local storage.

vojtechsimetka:
Resolves #48

### DIFF
--- a/src/providers/Theme.tsx
+++ b/src/providers/Theme.tsx
@@ -1,0 +1,55 @@
+import type { Theme } from "@material-ui/core/styles";
+import {createContext, ReactChild, useEffect, useState} from "react"
+import { lightTheme, darkTheme } from "../theme"
+
+interface ContextInterface {
+    themeMode: ThemeMode
+    theme: Theme
+    toggleThemeMode: () => void
+}
+
+const initialValues: ContextInterface = {
+    themeMode: 'light',
+    theme: lightTheme,
+    toggleThemeMode: () => {}
+}
+
+const Context = createContext<ContextInterface>(initialValues);
+
+interface Props {
+    children: ReactChild
+}
+
+function Provider({children}: Props) {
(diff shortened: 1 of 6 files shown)


matmertz25 (approved):
Thanks for adding this, works much better than a reload

vojtechsimetka:
Does it work for you? In my case it does not switch well for the second time. No idea why but I think we'll find out why after some more refactoring :)

matmertz25:
Yeah your right, it initially works for me but after flipping back and forth per your comment it is not transitioning from dark -> light correctly.

src/providers/Theme.tsx
```diff
@@ -0,0 +1,55 @@
...
+        localStorage.setItem('theme', newThemeMode)
+    }
+
+    // Detect previously store theme
+    useEffect(() => {
+        const newThemeMode: ThemeMode = localStorage.getItem('theme') === 'light' ? 'light' : 'dark'
+
+        setThemeMode(newThemeMode)
+
+        window?.matchMedia('(prefers-color-scheme: dark)')?.addEventListener('change', e => {
```
agazso, line 38: You pass an anonymous function to the `addEventListener` and another (but functionally same) anonymous function to `removeEventListener`. This is undefined (browser-dependent) behavior and usually it's better to extract the anonymous function and assign it to a variable, and pass the variable to the `addEventListener` and `removeEventListener` calls.

vojtechsimetka:
This should be looked at with fresh eyes. We probably don't need and want the provider altogether (as it is likely the reason why it does not work well) and get inspired by the official documentation https://material-ui.com/customization/palette/#user-preference

PR #50, opened 2021-04-02
feat: add pagination to tables

matmertz25:


vojtechsimetka:
@matmertz25 thanks for the PR!

 I'd like to understand a bit more why you decided we need to have pagination. AFAIK there is not that many peers your node would be connected to (<100) and having these paginated seems like a usability hinder to me (I don't know about you but I really hate clicking next page button instead of just scrolling :) ). Let me know what let you to do this change.

BTW it would be much easier to have some pagination function/hook which could be shared between different components. There are even such implementation on the internet and I think MUI has one as well. Something like:
```ts
interface Paginate<T> {
  currentPage: number
  currentPageData: T[]
  rowsPerPage: number
  setPage: (page: number) => {}
}
paginate<T[]>(data: T[], rowsPerPage: number): Paginate<T> => {
   ...
}
```

matmertz25:
I added pagination since I have a node that returns 829 balances in the balances tables, 112 cheques in the chequebook, and 112 settlements. It is my opinion that really any API that has the potential of returning 50+ records should either be paginated or virtualized for both performance and creating a more consolidated and organized UI. Even if we think some of these ranges wouldn't be too large the UI should be able to handle cases where the data is much larger; especially as the length of time of running a node increases. This could also accommodate a potential future where this UI could be used to manage a cluster of nodes.

As far as pagination of pages versus virtualization of the table, I dont mind changing to virtualization if that's what you would prefer. I completely understand the argument that scrolling tends to feel more natural than pages. I just added the pagination as a quick and easy feature to organize the UI for the first release. (in my case the node with 829 peer balances creates a very jarring UI).

Also, I like your suggestion with the hook as its much cleaner if we decide to do paginating.

PR #62, opened 2021-04-09
fix: file download redirects to bee node, upload limited to 1 file

### LINKED ISSUE #59: Download should just direct to new window with <BEE_NODE>/bzz/<hash>
It should also validate the hash with `Utils` from `bee-js` and show any issues to the user.

vojtechsimetka:
resolves #59 , #28 

- error handling for file search and upload
- upload limited to 1 file
- file is download by opening new window to the bee node

### DIFF
--- a/src/pages/files/index.tsx
+++ b/src/pages/files/index.tsx
@@ -44,22 +44,22 @@ export default function Files(): ReactElement {
 
   const [file, setFile] = useState<File | null>(null)
   const [uploadReference, setUploadReference] = useState('')
-  const [uploadingFile, setUploadingFile] = useState(false)
+  const [uploadError, setUploadError] = useState<Error | null>(null)
+  const [isUploadingFile, setIsUploadingFile] = useState(false)
 
   const uploadFile = () => {
     if (file === null) return
-    setUploadingFile(true)
+    setIsUploadingFile(true)
+    setUploadError(null)
     beeApi.files
       .uploadFile(file)
       .then(hash => {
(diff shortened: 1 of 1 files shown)


src/pages/files/index.tsx
```diff
@@ -38,131 +38,112 @@ export default function Files(): ReactElement {
 
   const [inputMode, setInputMode] = useState<'browse' | 'upload'>('browse')
   const [searchInput, setSearchInput] = useState('')
```
agazso, line 40: I am not sure that `searchInput` is the best name for this, maybe `hashInput` or `referenceInput` is more appropriate.

src/pages/files/index.tsx
```diff
@@ -38,131 +38,112 @@ export default function Files(): ReactElement {
...
+          Upload
+        </Button>
+      </div>
+      {inputMode === 'browse' && (
+        <>
+          <Paper className={classes.root}>
+            <InputBase
+              className={classes.input}
+              placeholder="Enter hash e.g. 0773a91efd6547c754fc1d95fb1c62c7d1b47f959c2caa685dfec8736da95c1c"
+              inputProps={{ 'aria-label': 'search swarm nodes' }}
```
agazso, line 106: I find this label a bit misleading, because it's somewhat true that this is what happens on the bee network, however from a user perspective here we want to fetch some content.

agazso (approved):
Looks good, but I haven't tested it.

Please check my comments regarding naming it as "search".

Issue #40, opened 2021-04-01
Bee and BeeDebug classes should be provided through context provider


vojtechsimetka:
Right now changing any of these URLs means the app needs to reload. This is not very user friendly so we should instead have both instances in a context provider. Instantiating it for each call is also unnecessary.

PR #63, opened 2021-04-09
feat: reviewed and simplified the node status check

vojtechsimetka:
Reviewed the wording and refactored to eliminate un-renderable states but also to make the node setup a bit cleaner. This can be even further simplified when we have a context provider (see #40 ) and we can even store state there.

### DIFF
--- a/src/pages/status/SetupSteps/ChequebookDeployFund.tsx
+++ b/src/pages/status/SetupSteps/ChequebookDeployFund.tsx
@@ -1,38 +1,22 @@
 import { Typography } from '@material-ui/core/'
-import { CheckCircle, Warning } from '@material-ui/icons/'
 import EthereumAddress from '../../../components/EthereumAddress'
 import DepositModal from '../../../components/DepositModal'
 import CodeBlockTabs from '../../../components/CodeBlockTabs'
-import type { ChequebookAddressResponse, ChequebookBalanceResponse } from '@ethersphere/bee-js'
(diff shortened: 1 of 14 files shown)


src/pages/status/SetupSteps/ChequebookDeployFund.tsx
```diff
@@ -1,38 +1,22 @@
...
+const ChequebookDeployFund = ({ isLoading, chequebookAddress, chequebookBalance }: Props): ReactElement | null => {
+  if (isLoading) return null
+
+  return (
+    <div>
+      <p style={{ marginBottom: '20px', display: 'flex' }}>
+        {chequebookAddress?.chequebookaddress && <DepositModal />}
+      </p>
+      <div style={{ marginBottom: '10px' }}>
+        {!(chequebookAddress?.chequebookaddress && chequebookBalance && chequebookBalance?.totalBalance > 0) && (
```
agazso, line 18: I find this logical condition a bit hard to follow. Maybe it would be worth extracting some values to variables to simplify.

agazso (approved):
Looks good, but I haven't tested 😄

src/pages/status/SetupSteps/ChequebookDeployFund.tsx
vojtechsimetka: Yeah these should be simple components that just display values that are always defined. But such rewrite seemed beyond the scope what we are doing now :)

PR #66, opened 2021-04-09
fix: withdraw/deposit converts from BZZ (10^16 base units)

### LINKED ISSUE #52: Withdraw/deposit is denominated in BZZ  digits
The withdraw and deposit modals use the smallest denomination of bzz and not 10^16 value. So inputing 1,000,000 means only 10^-10 BZZ will be transferred. This is not necessarilly wrong but should be clear to the users as the balances elsewhere in the app usually show the amount/10^16

vojtechsimetka:
Resolves #52


src/models/Token.ts
```diff
@@ -0,0 +1,58 @@
+import { BigNumber } from 'bignumber.js'
+import { isInteger, makeBigNumber } from '../utils'
+
+const BZZ_DECIMALS = new BigNumber(16)
+
+export class Token {
+  private amount: BigNumber
+  private readonly decimals: BigNumber
+
+  constructor(amount: BigNumber | string | bigint, decimals: BigNumber | string | bigint = BZZ_DECIMALS) {
```
agazso, line 10: I don't completely understand the reason that the `decimals` is not a constant (because it always used with the default value provided, except in tests) and even less why it is a BigNumber itself.

When checking the  code where it is used I only see that this is more likely a convenience functionality to avoid using very big numbers as the `amount`. However the maximum amount of decimals that can be used by the `bignumber.js` library seems to be limited to 1e9 according to the doc. 

https://mikemcl.github.io/bignumber.js/#pow-precision

It seems to me that by making `decimal` a constant could be less complex and error-prone to use while not losing functionality.

agazso (approved):
Looks good, please check my comment regarding the use of decimals.

src/models/Token.ts
vojtechsimetka: Thanks for pointing this out. I changed the decimals to be an integer value in a range 0-18

There is a bit of a misunderstanding here. There is the precision for the BigNumber itself and equivalent-ish precision for Tokens (Native, ERC20, ERC677...) . The Token class is aimed to represent any ethereum token which has a precision between 0 and 18 digits. This is just user facing simplification, because in solidity you only have integers. So lets have a token A with precision 3. When someone has 5.43 A tokens (this is what a Web3 wallet will show, any contract interaction is expecting it to be passed in as integer in what I call base units, so 5430.

If you use the BigNumber and set the pow precision to 3, it just means that anything beyond this precision is automatically rounded. So number `5.432` and `5.4321` are the same because the second one is rounded to 3 decimal places. This is a nice feature for doing operations with fixed precision, but not so nice when you are trying to prevent user passing in values that are out of range.

That is why I decided to use BigNumber without any precision limits and when user passes in decimal representation of a token A `5.4321` I multiply it by `10^3` and check if the result is integer. If it is not integer it means that the input user puts in for withdrawing/depositing is out of precision range and would have to be rounded/truncated. IMHO that would most likely indicate a typo.

PR #69, opened 2021-04-15
feat: cashing out for each peer and single page accounting

vojtechsimetka:
Flatten the whole accounting into single page. In addition to balance and settlement totals, it retrieves and calculate the `uncashedAmount` for each each peer and if non-zero it shows `cashout` button.

Screenshot 2021-04-16 at 20 18 44


src/pages/accounting/BalancesTable.tsx
```diff
@@ -1,79 +1,90 @@
...
+      <Table className={classes.table} size="small" aria-label="simple table">
+        <TableHead>
+          <TableRow>
+            <TableCell>Peer</TableCell>
+            <TableCell style={{ textAlign: 'right' }}>Outstanding Balance</TableCell>
+            <TableCell style={{ textAlign: 'right' }}>Settlements Sent / Received</TableCell>
+            <TableCell style={{ textAlign: 'right' }}>Total</TableCell>
+            <TableCell style={{ textAlign: 'right' }}>Uncashed Amount</TableCell>
+            <TableCell style={{ textAlign: 'right' }}></TableCell>
+          </TableRow>
```
Cafe137, line 39: The duped inline styles look a bit off here. Can they be replaced with a `TableHead TableCell :not-first` or similar rule?

src/pages/accounting/BalancesTable.tsx
```diff
@@ -1,79 +1,90 @@
...
-                  <TableCell align="right"></TableCell>
-                </TableRow>
-              ))}
-            </TableBody>
-          </Table>
-        </TableContainer>
-      )}
-    </div>
+    <TableContainer component={Paper}>
+      <Table className={classes.table} size="small" aria-label="simple table">
```
Cafe137, line 30: Not sure what the approach on a11y is, but maybe if this is specified here, then it could read `Balances Table`

src/pages/accounting/AccountCard.tsx
```diff
@@ -40,81 +40,60 @@ interface ChequebookBalance {
...
+              <Typography component="p" variant="h5">
+                {fromBZZbaseUnit(chequebookBalance?.availableBalance || 0).toFixed(7)} BZZ
+              </Typography>
+            </div>
+            <div>
+              <Typography component="h2" variant="h6" color="primary" gutterBottom>
+                Total Sent / Received
+              </Typography>
+              <Typography component="div" variant="h5">
+                <span style={{ marginRight: '7px' }}>
```
Cafe137, line 86: `7px` stands out to me, since `theme` has spacings, also it is not an increment of 4 or 8 units (not sure if we follow that here).

src/pages/accounting/BalancesTable.tsx
```diff
@@ -1,79 +1,90 @@
...
+              <TableCell className={classes.values}>
+                -{fromBZZbaseUnit(sent).toFixed(7)} / {fromBZZbaseUnit(received).toFixed(7)} BZZ
+              </TableCell>
+              <TableCell className={classes.values}>
+                <span
+                  style={{
+                    color: balance + received - sent > 0 ? '#32c48d' : '#c9201f',
+                  }}
+                >
+                  {fromBZZbaseUnit(balance + received - sent).toFixed(7)}
```
Cafe137, line 71: Maybe the expression `balance + received - sent` could be extracted as a variable

src/pages/accounting/BalancesTable.tsx
vojtechsimetka: I'd prefer to keep this configurable on a per column basis. But I will simplify to use the `align="right"`

src/pages/accounting/AccountCard.tsx
vojtechsimetka: Yeah I am removing it all together as the margin is not needed at all now thanks to the `columnGap` spacing.

Cafe137 (approved):

src/components/CashoutModal.tsx
```diff
@@ -67,32 +71,33 @@ export default function DepositModal(): ReactElement {
...
-              onChange={e => setPeerId(e.target.value)}
-            />
-          </DialogContent>
-        )}
+        <DialogContent>
+          <DialogContentText style={{ marginTop: '20px', overflowWrap: 'break-word' }}>
+            {loadingCashout && (
+              <>
+                <span>
+                  Cashing out <strong>{fromBZZbaseUnit(uncashedAmount).toFixed(7)}</strong> from Peer{' '}
```
agazso, line 79: It would be nice to use constants instead of number literals, especially if they are repeated.

src/hooks/accounting.ts
```diff
@@ -1,57 +1,87 @@
...
-  const settlements = useApiSettlements()
-  const balances = useApiPeerBalances()
   const accounting: Record<string, Accounting> = {}
-  const [isLoadingUncashed, setIsloadingUncashed] = useState<boolean>(false)
-  const [uncashedAmounts, setUncashedAmounts] = useState<CashoutInfo[] | null>(null)
 
-  balances.peerBalances?.balances.forEach(
-    // Some peers may not have settlement but all have balance (therefore initialize sent + received to 0)
-    ({ peer, balance }) => (accounting[peer] = { balance, sent: 0, received: 0, uncashedAmount: 0 }),
+  balances.forEach(
```
agazso, line 34:
```suggestion
  balances?.forEach(
```

src/hooks/accounting.ts
```diff
@@ -1,57 +1,87 @@
...
+    // Some peers may not have settlement but all have balance (therefore initialize sent, received and uncashed to 0)
+    ({ peer, balance }) => (accounting[peer] = { peer, balance, sent: 0, received: 0, uncashedAmount: 0 }),
   )
 
-  settlements.settlements?.settlements.forEach(
-    ({ peer, sent, received }) => (accounting[peer] = { ...accounting[peer], sent, received }),
+  settlements.forEach(({ peer, sent, received }) => (accounting[peer] = { ...accounting[peer], sent, received }))
```
agazso:
```suggestion
  settlements?.forEach(({ peer, sent, received }) => (accounting[peer] = { ...accounting[peer], sent, received }))
```

src/components/CashoutModal.tsx
vojtechsimetka: Apologies, what do you mean here? Extract the `fromZZZBaseUnit(...)` into the hooks directly and just display it here? So it would be:

```suggestion
                  Cashing out <strong>{uncashedAmount}</strong> from Peer{' '}
```

src/pages/accounting/BalancesTable.tsx
```diff
@@ -34,10 +35,11 @@ function BalancesTable({ accounting, isLoadingUncashed }: Props): ReactElement |
             <TableCell style={{ textAlign: 'right' }}>Settlements Sent / Received</TableCell>
             <TableCell style={{ textAlign: 'right' }}>Total</TableCell>
             <TableCell style={{ textAlign: 'right' }}>Uncashed Amount</TableCell>
+            <TableCell style={{ textAlign: 'right' }}></TableCell>
           </TableRow>
         </TableHead>
         <TableBody>
-          {Object.entries(accounting).map(([peer, values]) => (
+          {accounting.map(({ peer, balance, received, sent, uncashedAmount }) => (
```
agazso:
```suggestion
          {accounting?.map(({ peer, balance, received, sent, uncashedAmount }) => (
```

src/hooks/accounting.ts
vojtechsimetka: No need, the balances can not be null here.

src/pages/accounting/BalancesTable.tsx
vojtechsimetka: It can not be null here, it is already checked for null previously.

src/hooks/accounting.ts
vojtechsimetka: The settlements are checked earlier: `if (!balances || !settlements) return null`

src/pages/accounting/BalancesTable.tsx
```diff
@@ -50,29 +52,33 @@ function BalancesTable({ accounting, isLoadingUncashed }: Props): ReactElement |
...
+                    color: balance + received - sent > 0 ? '#32c48d' : '#c9201f',
                   }}
                 >
-                  {fromBZZbaseUnit(values.balance + values.received - values.sent).toFixed(7)}
+                  {fromBZZbaseUnit(balance + received - sent).toFixed(7)}
                 </span>{' '}
                 BZZ
               </TableCell>
               <TableCell className={classes.values}>
                 {isLoadingUncashed && 'loading...'}
```
agazso, line 76: I can imagine this expressions expressed as
```tsx
{isLoadingUncashed 
 ? 'loading'
 : uncashedAmount > 0 
    ? <>{fromBZZbaseUnit(uncashedAmount).toFixed(7)} BZZ</>
    : '0 BZZ'
```

Or break them down to smaller components (e.g. Loading/UncashedAmount)

agazso (approved):
Looks good, but please check out the optional chaining operators. I only reviewed this on Github but it would be good if there would be compiler warnings when they are missing because we are potentially referencing null/undefined values.

src/components/CashoutModal.tsx
agazso: I mean instead of using `.toFixed(7)` it would be better to use `.toFixed(BZZ_DIGITS)` (or whatever name makes sense)

src/pages/accounting/BalancesTable.tsx
vojtechsimetka: Thanks, I decided to go half way as nested ternary operators are tough to read. I agree this should be done as a separate component but I know of few other places where that is the case and would love to address that separately.
```tsx
{isLoadingUncashed && 'loading...'}
{!isLoadingUncashed && <>{uncashedAmount > 0 ? fromBZZbaseUnit(uncashedAmount).toFixed(7) : '0'} BZZ</>}
```

Issue #72, opened 2021-04-21
Add Dockerfile support


tmm360:
Add a Dockerfile for support container publication.

PR #75, opened 2021-04-21
feat: added Dockerfile (#72)

tmm360:


vojtechsimetka:
This looks good to me, WDYT @nugaon ?

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:15.14-alpine AS build
+WORKDIR /src
+COPY . .
+RUN npm ci
+RUN npm run build
+
+FROM node:15.14-alpine as final
+RUN yarn global add serve
```
nugaon, line 8: why do we use `yarn` here instead of `npm`?

nugaon: This whole setup could happen by running only `npx serve` instead. I don't see justified this additional layer.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:15.14-alpine AS build
+WORKDIR /src
+COPY . .
+RUN npm ci
+RUN npm run build
+
+FROM node:15.14-alpine as final
```
nugaon, line 7: let's be unified 
```suggestion
FROM node:15.14-alpine AS final
```

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN yarn global add serve
+WORKDIR /app
+COPY --from=build /src/build .
+EXPOSE 5000
```
nugaon, line 11: I can't see anywhere it would run on different port than the default `3000`, or am I wrong?

tmm360: It's exposed on 5000

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ENTRYPOINT ["serve"]
```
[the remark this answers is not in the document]
tmm360: It works also without to me, but can add, no problem

Dockerfile
tmm360: I'm following the readme instructions. Will try to use `npx serve`

nugaon: > The quickest way to get started is to just run npx serve in your project's directory.
https://www.npmjs.com/package/serve#usage

This is the first sentence in the readme instructions, after that just an other way is explained how to do it.

tmm360: Running with `npx serve` it says:
`npm WARN exec The following package was not found and will be installed: serve`

So it's to choose if install it with container, or with first execution. My opinion, best to include before. A container should work as is, minimizing requirements to external resources to run. More predictable.

nugaon: yeah I see now, you have only the `build` folder. you are right, I just wanted to propose to use only one package manager in the same workflow, can you try with this please?
```suggestion
RUN npm i -g serve
```

vojtechsimetka: Thanks @tmm360 and @nugaon , we just merged a PR which adds serve as devDependency so that users who want to use it directly outside of Docker have simpler life. I believe it also solves this discussion. Cheers :) https://github.com/ethersphere/bee-dashboard/pull/79

tmm360: @vojtechsimetka Unfortunately because `serve` is imported as a dev dependency, and I'm optimizing image exporting only build result, serve isn't available on the final image.
Instead "RUN npm i -g serve" works. At this point I'm going to take this last one.

vojtechsimetka:
I am very sorry @tmm360 but we are making yet another changes how the app is served. See #88 . I presume this might affect the PR.

tmm360:
> I am very sorry @tmm360 but we are making yet another changes how the app is served. See #88 . I presume this might affect the PR.

Probably yes, I will wait for #88 to be merged for update this. Thanks

tmm360:
The built version inside `./build` directory doesn't report the `package.json` file, required by the `npm run serve` command, so for Dockerfile I still use the `serve` command installed globally.

Anyway the `serve` command exposes by default on port `5000`, so I've updated it to `8080` for standardize with documentation.

I've also updated readme instructions for a local build with Docker, until image will be published on docker hub.

README.md
```diff
@@ -45,6 +45,15 @@ npm run serve
...
 You can now access Bee Dashboard on [http://localhost:8080/](http://localhost:8080/)
 
+## Docker
+
+From inside project directory:
+
+```sh
+docker build . -t bee-dashboard
+docker run -p 8080:8080 bee-dashboard
+```
```
nugaon, line 55:
```suggestion
To build Docker image, execute the following from inside project directory:

```sh
docker build . -t bee-dashboard
docker run --rm -p 127.0.0.1:8080:8080 bee-dashboard
```
```

nugaon (approved):
LGTM, Thanks for the contrib! : )

after merging this, we can create the corresponding CI PR to publish this image under `ethersphere` IMO

README.md
tmm360: Missing the code block closing marker ```

vojtechsimetka: It's not, but github thinks it is the opening of next codeblock :/

PR #76, opened 2021-04-21
fix: adjusted prettier "endOfLine" to "auto" (#73)

### LINKED ISSUE #73: Change prettier endOfLine to "auto"
Currently in file `.prettierrc` is set as
```
{
  ...
  "endOfLine": "lf",
  ...
}
```
and this prevent to compile with windows.

Instead it should be set `"endOfLine": "auto"`.

tmm360:
Fixes #73

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,7 +7,7 @@
   "singleQuote": true,
   "quoteProps": "as-needed",
   "trailingComma": "all",
-  "endOfLine": "lf",
+  "endOfLine": "auto",
   "arrowParens": "avoid",
   "proseWrap": "always"
 }


vojtechsimetka:
Hi @tmm360 thanks for this PR, I've asked @nugaon if he can review as this is not my expertise :)

nugaon:
It seems to me an outdated config value as it says in here https://prettier.io/docs/en/options.html#end-of-line

tmm360:
I'm using Windows. For avoid any problem with type of end line into repos, I'm using the git global config `core.autocrlf=true`, that automatically convert into CRLF when checkout, and into LF when commit.
When I use Docker for build projects, it runs into a Linux kernel with WSL, but takes sources from Windows's FS where I've checked out, so importing files converted with CRLS. Prettier with `"ls"` config makes it to fail. Instead setting it to `"auto"` permits it to succeeded also in this case, without any disadvantages.
I don't understand the decision to set `"ls"` as default, but it isn't an outdated config, in my opinion.

nugaon:
Where can I find this setup which assures all the commit code will use `lf` instead of `crlf`?

I don't really get this global config of yours because in theory (I'm not a Windows user) any Windows editor should handle this line ending.

> All modern text editors in all operating systems are able to correctly display line endings when \n (LF) is used. However, old versions of Notepad for Windows will visually squash such lines into one as they can only deal with \r\n (CRLF).

Is it possible to change your configuration in that way this pettier config wouldn't cause this problem?

tmm360:
> Where can I find this setup which assures all the commit code will use lf instead of crlf?

https://git-scm.com/book/it/v2/Customizing-Git-Git-Configuration under `core.autocrlf`

> I don't really get this global config of yours because in theory (I'm not a Windows user) any Windows editor should handle this line ending.
> Is it possible to change your configuration in that way this pettier config wouldn't cause this problem?

It's the default configuration of git with Windows. Obviously could be changed, never changed because never needed.

nugaon:
> It's the default configuration of git with Windows. Obviously could be changed, never changed because never needed.

Can you please change then your config to be aligned with our setup?

tmm360:
> Can you please change then your config to be aligned with our setup?

Of course I can.

> That's why I ask this from you because we want unified line endings at all of our repos/source files and this client side config of yours doesn't justify the change

Because of this setting I only push code with LF ending. Anyway, I can set `core.autocrlf=input` so at checkout is kept LF, and any eventual new file introduced with CRLF is converted to LF on commit.

PR #85, opened 2021-04-27
fix: if any error occurred, the first step could not be expanded

vojtechsimetka:


src/pages/status/NodeSetupWorkflow.tsx
```diff
@@ -95,9 +95,13 @@ export default function NodeSetupWorkflow({
...
-    if (activeStep > 0 && activeStep < steps.length) return
+    if (activeStep >= 0 && activeStep < steps.length) return
+
+    // If any step is not fully loaded yet return
+    if (!steps.every(step => !step.isLoading)) return
 
     // Select first step that is not OK
+    // This is deliberately a for condition so that we can terminate the useEffect from within the cycle
     for (let i = 0; i < steps.length; ++i) {
       if (!steps[i].condition) {
```
agazso, line 106: It would be better to call the `condition` `isOk` instead (as it is called in the `StatusHookCommon`).

agazso (approved):

src/pages/status/NodeSetupWorkflow.tsx
vojtechsimetka: Great suggestion

Issue #98, opened 2021-05-17
run serve failed


rickiey:
> @ethersphere/bee-dashboard@0.1.0 serve /root/workplace/bee-dashboard
> http-serve -o ./build

Starting up http-serve for ./public
Available on:
  http://127.0.0.1:8080
  http://myIP:8080
  http://172.16.0.147:8080
  http://172.16.0.251:8080
Hit CTRL-C to stop the server
[2021-05-17T06:09:24.658Z] "GET /%PUBLIC_URL%/favicon.ico" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"
(node:12164) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
(Use `node --trace-deprecation ...` to show where the warning was created)
[2021-05-17T06:09:25.060Z] "GET /%PUBLIC_URL%/favicon.ico" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"
[2021-05-17T06:09:26.897Z] "GET /%PUBLIC_URL%/manifest.json" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"
[2021-05-17T06:09:28.029Z] "GET /" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"
[2021-05-17T06:09:28.214Z] "GET /%PUBLIC_URL%/favicon.ico" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"
[2021-05-17T06:09:28.674Z] "GET /%PUBLIC_URL%/favicon.ico" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36"

Request URL: http://myIP:8080/%PUBLIC_URL%/favicon.ico
Request Method: GET
Status Code: 400 Bad Request

vojtechsimetka:
Thanks for reporting, will look into it.

Issue #96, opened 2021-05-13
The webserver does not redirect to `/` when refreshing page


vojtechsimetka:
The serve system does not serve the website as single page app. So refreshing on some page like `/accounting` gives error `404`.

PR #101, opened 2021-05-18
fix: serve npm command path specification

agazso:
When running the `http-serve -o ./build` command the path is not changed and it uses the default value (`./public`).

This seems to be a bug in `http-serve`. This also explains #98

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "serve": "http-serve -o ./build",
+    "serve": "http-serve ./build -o",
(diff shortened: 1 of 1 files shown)


vojtechsimetka (approved):

matmertz25 (approved):
I noticed this as well, I tried also adding the '--fallback' parameter to also get the routing to work on refresh (issue #96) but couldn't get it to work https://github.com/dkarmalita/http-serve#available-options. Did you try adding the fallback option?

matmertz25:
I think it may not be working correctly with the http-serve package because the application currently uses paths with a trailing slash (I.e. /accounting/) instead of just /accounting but haven't had a chance to test this out.

vojtechsimetka:
> I think it may not be working correctly with the http-serve package because the application currently uses paths with a trailing slash (I.e. /accounting/) instead of just /accounting but haven't had a chance to test this out.

Yeah, I think we should return back to the [serve](https://www.npmjs.com/package/serve) package

Issue #112, opened 2021-05-25
Feature Request: Add option to change port


Zaphod101010:
Port 8080 is hard coded.  I already have another server running on port 8080.

I'd like to be able to manually change the port bee-dashboard runs on.

It would be great if it could also do some checks on startup to see if port 8080 is already in use, then increment the port by 1 until it finds an available port.

vojtechsimetka:
Thanks @Zaphod101010 for the suggestion. It makes sense but is not currently our priority. If anyone wants to try and implement it I'll promise to review timely :).

In the meantime, if someone wants to run the project on a different port, the easiest way is to do following is to download the files from git, build the project and serve it with any http server.

```
git clone git@github.com:ethersphere/bee-dashboard.git
cd bee-dashboard
npm ci
npm run build
```

And serve the `./build` folder with any http server you want. You can also use the `http-serve` we are using, just edit the configuration https://github.com/ethersphere/bee-dashboard/blob/aab0462047a3fcd87ba258b5486aede922865b1e/serve.js#L12-L16

PR #114, opened 2021-05-26
Find free port to use during startup

Zaphod101010:
Maybe something like this to automatically find the first open port above 8080?

### DIFF
--- a/getPort.js
+++ b/getPort.js
@@ -1,0 +1,89 @@
+
+/*
+ * 
+ *  Starting at port 8080, increment port by one until a free port is found
+ *
+ */
+
+async function getAvailablePort() {
+  const port = 8080;
+  for await ( const result of checkPorts(port) ) {
+    if (result.available) {
+      return result.port;
+    }
+  }
+}
+
+async function* checkPorts(port) {
+  let available;
+  while (!available) {
+    if (['linux', 'darwin'].includes(process.platform)) {
+      available = await checkPortMacLinux(port).catch((err) => console.error(err));
+    } else {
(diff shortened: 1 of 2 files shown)


vojtechsimetka:
Hi @Zaphod101010 thank you very much for this contribution. I truly appreciate it.

I am sorry I was not clear, my request was more about the part where users can supply their own preferred port on which the app should run (as per the title of the issue #112). So user could do `bee-dashboard serve --port xxxx`.

We talked about your PR within the team and we are not fans of the app automatically starting on a random port on its own (if 8080 is busy).  Instead we will probably choose different statistically less utilised port. That being said, I could see this check and suggestion of different port to be useful but the user always has to be in charge so this has to be interactive (and disabled in CI). So similar to how the `react-script`'s `start` command works, it could detect the port is busy and suggest different port to run the app on. Only after confirmation from user this port would be used. I would also prefer if such functionality was not implemented by us and instead some existing npm module was used (e.g. the one used in `react-scripts` - https://www.npmjs.com/package/detect-port-alt )

Zaphod101010:
The problem with solutions like the one in package you linked is that it checks for available ports by trying to listen on that port, then checking for an error.  That check fails when the port binding looks like this:

```
lnd      1496 zaphod   23u  IPv4   TCP *:http-alt (LISTEN)
node    19264 zaphod   21u  IPv4   TCP localhost:http-alt (LISTEN)
```

Node and LND are both running on port 8080.

Using the OS level commands are a much more solid way to see what is actually running on the computer.

PR #115, opened 2021-05-26
feat: postage stamps support

vojtechsimetka:
https://user-images.githubusercontent.com/7974813/119629116-005f5780-be0e-11eb-969c-7df0cd55c8a9.mov


agazso (approved):
The code looks good.

My only concern is that when buying a postage stamp fails (error 400, "out of funds") then there is no error indication whatsoever and the dialog just stays there. This could be probably addressed in a different PR though.

PR #122, opened 2021-05-28
fix: replace http-serve with serve package

### LINKED ISSUE #96: The webserver does not redirect to `/` when refreshing page

### LINKED ISSUE #106: After updating to latest version from npmjs, browser displays cached version
I installed globally the `0.2.0` version and when I run it, browser still served me the `0.1.0` version from cache. Hard refresh fixed this, but I'm sure it will be common issue.

We should add cache invalidation to the serve system command

matmertz25:
Replaces the http-serve package to properly handle the single page routing. 

- Resolves #96 to handle refreshes on the single page application
- Resolves #106 with cache invalidation
- Half addresses #112 as the `serve` package will select an alternative port other than 8080 if it is not available.


vojtechsimetka:
Thank @matmertz25 , this is great you are doing exactly what I intended here. Please do solve the `serve` command in `bin` as well though (that is what makes the `bee-dashboard` work).

https://github.com/ethersphere/bee-dashboard/pull/122/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R16

liuxingbaoyu:
Great work!
This should fix them too.

https://github.com/ethersphere/bee-dashboard/issues/116
https://github.com/ethersphere/bee-dashboard/issues/97
https://github.com/ethersphere/bee-dashboard/issues/86
https://github.com/ethersphere/bee-dashboard/issues/108

This should be the same problem we encountered. I tested this commit to fix it.
https://github.com/http-party/http-server/issues/525

liuxingbaoyu:
Also on my computer (windows 10), `serve build --single --listen 8080` will use `Windows Script` to run `.\serve.js` instead of `.\node_modules\.bin\serve.cmd`.

serve.js
```diff
@@ -1,16 +1,34 @@
...
-server.listen(8080, '127.0.0.1', function () {
+  return handler(request, response, serverConfig);
+})
+ 
+server.listen(8080, () => {
   console.log('Starting up Bee Dashboard on address http://localhost:8080')
   console.log('Hit CTRL-C to stop the server')
   opener('http://localhost:8080')
-})
+});
```
vojtechsimetka, line 34:
```suggestion
})

```

public/serve.json
```diff
@@ -0,0 +1,12 @@
...
+    "headers": [
+        {
+          "source" : "*",
+          "headers" : [{
+            "key" : "Cache-Control",
+            "value" : "max-age=3600"
+          }]
+        }
+    ]    
+}
```
vojtechsimetka, line 12:
```suggestion
}

```

package.json
```diff
@@ -70,7 +71,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "serve": "http-serve ./build -o",
+    "serve": "serve build --single --listen 8080",
```
vojtechsimetka, line 74: Do we actually need the `serve` package and to run it here in this way instead of `node ./serve.js`? Asking because we encourage the users to install the tool globally with `npm i @ethersphere/bee-dashboard -g` and to run it as `bee-dashboard serve`. That means they actually run the `serve` command in the `bin` part of this file and not this one defined in `scripts`. Could we remove the `serve` package and unify? (I am asking as I genuinely don't know)

https://github.com/ethersphere/bee-dashboard/pull/122/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R16

vojtechsimetka:
Oh also could you please commit the `package-lock.json` file as well? We usually install dependencies with `npm ci` which actually takes the packages definition from the lock file and not the `package.json`.

package.json
matmertz25: I dont think we need the serve package but we would still need the core dependency of `serve-handler` to run the serve.js file. We can pull it out if you would like as like you said we dont really need it in the current capacity that we run the dashboard.

vojtechsimetka: I would actually prefer that. So let's add the `serve-handler`, remove the `serve` dependency and change this line to `"serve": "node ./serve.js"`

BTW it is always recommended to add any dependencies one uses for two reasons:
- at some point we may decide to remove the package that includes it as a dependency and therefore break it
- the package that includes it as a dependency gets updated and removes it

This reminds me we should add `depcheck` that will catch these things.

vojtechsimetka (approved):

PR #149, opened 2021-06-18
feat: display postage batch usage percentage

Cafe137:
image

image


src/providers/Stamps.tsx
```diff
@@ -29,8 +34,21 @@ interface Props {
   children: ReactChild
 }
 
+function enrichStamp(postageBatch: PostageBatch): EnrichedPostageBatch {
+  const { depth, bucketDepth, utilization } = postageBatch
+
+  const usage = utilization / Math.pow(2, depth - bucketDepth)
+  const usageText = Math.ceil(usage * 100) + '%'
```
vojtechsimetka, line 41: I'd prefer a template here to avoid mistakes in future.

```suggestion
  const usageText = `${Math.ceil(usage * 100)} %`
```

vojtechsimetka (approved):

PR #165, opened 2021-08-10
feat: synchronized platform tabs

### LINKED ISSUE #10: Troubleshooting OS tabs should be syncronized
Changing tab from `Linux` to `macOS` should alter all the tab instances.

Screenshot 2021-03-10 at 17 17 00

### LINKED ISSUE #65: AntTabs should be extracted into reusable component
They are right now defined inside of two components https://github.com/ethersphere/bee-status/search?q=antTab

- [x] AccountingPage https://github.com/ethersphere/bee-status/blob/master/src/pages/accounting/index.tsx
- [ ] CodeBlocksTabs https://github.com/ethersphere/bee-status/blob/master/src/components/CodeBlockTabs.tsx

Should be extracted out and reused.

vojtechsimetka:
Resolves #10
Fixes #65

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,7 +9,7 @@ import { SnackbarProvider } from 'notistack'
 import BaseRouter from './routes/routes'
 import { lightTheme, darkTheme } from './theme'
 import { Provider as StampsProvider } from './providers/Stamps'
-import { Provider as PlatfomrProvider } from './providers/Platform'
+import { Provider as PlatformProvider } from './providers/Platform'
 
 const App = (): ReactElement => {
   const [themeMode, toggleThemeMode] = useState('light')
@@ -37,7 +37,7 @@ const App = (): ReactElement => {
     <div className="App">
       <ThemeProvider theme={themeMode === 'light' ? lightTheme : darkTheme}>
         <StampsProvider>
-          <PlatfomrProvider>
+          <PlatformProvider>
             <SnackbarProvider>
               <>
                 <CssBaseline />
@@ -46,7 +46,7 @@ const App = (): ReactElement => {
                 </Router>
               </>
             </SnackbarProvider>
-          </PlatfomrProvider>
+          </PlatformProvider>
         </StampsProvider>
       </ThemeProvider>
     </div>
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -9,6 +9,7 @@ import { SnackbarProvider } from 'notistack'
 import BaseRouter from './routes/routes'
 import { lightTheme, darkTheme } from './theme'
 import { Provider as StampsProvider } from './providers/Stamps'
+import { Provider as PlatfomrProvider } from './providers/Platform'
```
Cafe137, line 12: Typo :)

src/providers/Platform.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export function Provider({ children }: Props): ReactElement {
+  const [platform, setPlatform] = useState<SUPPORTED_PLATFORMS>(SUPPORTED_PLATFORMS.Linux)
+
+  // This is in useEffect as it really just needs to run once and not on each re-render
+  useEffect(() => {
+    const os = getOS()
+
+    setPlatform(isSupportedPlatform(os) ? os : SUPPORTED_PLATFORMS.Linux)
```
Cafe137, line 64: I find the Linux fallback a bit confusing.

My understanding is that it is used only for the troubleshooting, so it does not cause any issues right now, but in the future it could be a surprise that unsupported platforms just fall back to Linux. I would expect `platform` to always reflect my OS correctly.

Even now, an `if (SUPPORTED_PLATFORMS[platform] === undefined)` could show a warning on the troubleshooting screen maybe. Or it could be extracted as a `supported: boolean` to the Context.

Just some food for thought, I am not entirely convinced this should be refactored right away. 🤔

Cafe137: Oh nvm, I think this is for the default value only. 🙂

Cafe137 (commented):
LGTM - let me know if you had the chance to try out Viktor's enum refactor idea. 🙂

src/providers/Platform.tsx
vojtechsimetka: Yes indeed :)

vojtechsimetka:
> LGTM - let me know if you had the chance to try out Viktor's enum refactor idea. 🙂

I decided not to implement it in the end because it does not solve the index issue. There would have to be additional logic in the provider and the tab component which just did not seem worth it and would make this way more complicated.

Cafe137 (approved):

PR #166, opened 2021-08-11
feat: add retry to accounting

Cafe137:
Reworked `Promise.all` to `Promise.allSettled`

Added utility fn to create retriable Promise

Before:

Screenshot 2021-08-11 at 14 00 58

After:

Screenshot 2021-08-11 at 14 02 44

### DIFF
--- a/src/hooks/accounting.ts
+++ b/src/hooks/accounting.ts
@@ -2,6 +2,7 @@ import { LastCashoutActionResponse } from '@ethersphere/bee-js'
 import { useEffect, useState } from 'react'
 import { Token } from '../models/Token'
 import { beeDebugApi } from '../services/bee'
+import { makeRetriablePromise, unwrapPromiseSettlements } from '../utils'
 import { Balance, Settlement, useApiPeerBalances, useApiSettlements } from './apiHooks'
 
 interface UseAccountingHook {
@@ -80,11 +81,10 @@ export const useAccounting = (): UseAccountingHook => {
   const settlements = useApiSettlements()
   const balances = useApiPeerBalances()
 
-  const [err, setErr] = useState<Error | null>(null)
   const [isLoadingUncashed, setIsloadingUncashed] = useState<boolean>(false)
   const [uncashedAmounts, setUncashedAmounts] = useState<LastCashoutActionResponse[] | undefined>(undefined)
 
-  const error = balances.error || settlements.error || err
+  const error = balances.error || settlements.error
 
   useEffect(() => {
     // We don't have any settlements loaded yet or we are already loading/have loaded the uncashed amounts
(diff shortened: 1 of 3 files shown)


Cafe137:
There was a bug in the retry logic - unsuccessful retries resulted in pending promises instead of rejected.

After the fix, the table loads even if a request failed all three times.

Introduces a known issue: the uncashed amount is 0 in the previous rejected scenario. It would be much better to write 'N/A' or 'Could not fetch data', or a warning symbol with 'Something went wrong, check Bee logs' tooltip, you get the idea.

Not sure if it is needed though, my assumption is that the table does not fail too often, and this PR already mitigates those rare cases. 🙂

src/utils/index.ts
```diff
@@ -32,3 +32,59 @@ export function makeBigNumber(value: BigNumber | BigInt | number | string): BigN
...
+
+export async function sleepMs(ms: number): Promise<void> {
+  await new Promise<void>(resolve =>
+    setTimeout(() => {
+      resolve()
+    }, ms),
+  )
+}
+
+// TODO doc
```
vojtechsimetka, line 54: Could you please add the documentation or at least remove the todo? :)

src/utils/index.ts
```diff
@@ -32,3 +32,59 @@ export function makeBigNumber(value: BigNumber | BigInt | number | string): BigN
...
+export function mapPromiseSettlements<T>(promises: PromiseSettledResult<T>[]): PromiseSettlements<T> {
+  const fulfilled = promises.filter(promise => promise.status === 'fulfilled') as PromiseFulfilledResult<T>[]
+  const rejected = promises.filter(promise => promise.status === 'rejected') as PromiseRejectedResult[]
+
+  return { fulfilled, rejected }
+}
+
+// TODO doc
```
vojtechsimetka, line 62: Could you please add the documentation? :)

src/hooks/accounting.ts
```diff
@@ -93,12 +93,13 @@ export const useAccounting = (): UseAccountingHook => {
...
     const promises = settlements.settlements.settlements
       .filter(({ received }) => received.toBigNumber.gt('0'))
-      .map(({ peer }) => beeDebugApi.chequebook.getPeerLastCashout(peer))
+      .map(({ peer }) => makeRetriablePromise(() => beeDebugApi.chequebook.getPeerLastCashout(peer)))
 
-    Promise.all(promises)
-      .then(setUncashedAmounts)
-      .catch(setErr)
-      .finally(() => setIsloadingUncashed(false))
+    Promise.allSettled(promises).then(settlements => {
```
vojtechsimetka, line 98: So with the retries, this means that if there is repeatedly error, we have no results until all the retries are exhausted, right?

I am fine with it, just wanted to give you a food for thought.

vojtechsimetka (commented):
Kudos for making the `makeRetriablePromise` as iteration and not recursive.

src/utils/index.ts
Cafe137: Added :)

src/utils/index.ts
Cafe137: Added :)

src/hooks/accounting.ts
Cafe137: Yes, that is correct.

I am going to be spend at most an hour on making this a bit smoother, but if it is not doable in an easy way, I'd say it is not a big loss. 🙂

vojtechsimetka: I agree, thanks!

vojtechsimetka (approved):
Thank you, this is great!

PR #169, opened 2021-08-12
feat: add tooltips and health indicator to peers

Cafe137:
Screenshot 2021-08-12 at 15 42 04
Screenshot 2021-08-12 at 15 42 12
Screenshot 2021-08-12 at 15 42 37


Cafe137:
@agazso @vojtechsimetka What do you think of this approach? I think optimal values can be easily grabbed from Discord, but I am having trouble writing meaningful and useful explanations for these cases.

Cafe137:
(The Overall Health Indicator is completely static and WIP yet)

Cafe137:
Another potential issue here is the ever-changing nature of the stats. Population may grow or drop rapidly, max connections may be extended, not sure about depth. Do we want to keep updating them manually?

vojtechsimetka:
So we are actually going to do an auto refresh every ~30 seconds just as we do with the stamps.

Cafe137:
> So we are actually going to do an auto refresh every ~30 seconds just as we do with the stamps.

Sorry, I meant to optimal values

2 weeks later a healthy population may be 200K, not 100K

agazso:
> @agazso @vojtechsimetka What do you think of this approach? I think optimal values can be easily grabbed from Discord, but I am having trouble writing meaningful and useful explanations for these cases.

I think the direction is good, we discussed something similar with @vojtechsimetka already. Maybe I would prefer less text on the screen, one idea could be to show a label and when you hover it shows a tooltip or displays the explanation.

https://onaircode.com/html-css-tooltip-examples/

agazso:
> > So we are actually going to do an auto refresh every ~30 seconds just as we do with the stamps.
> 
> Sorry, I meant the optimal values
> 
> 2 weeks later a healthy population may be 200K, not 100K

I would not worry about that, hopefully the numbers only go up and with an updated version we can update the numbers as well.

Cafe137:
Screenshot 2021-08-13 at 11 51 41
Screenshot 2021-08-13 at 11 54 32
Screenshot 2021-08-13 at 11 55 03

src/components/Title.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <RowBetween>
+          <Typography className={classes.title} color="textSecondary" gutterBottom>
+            {label}
+          </Typography>
+          <Info />
+        </RowBetween>
+      </span>
+    </Tooltip>
+  )
+}
```
vojtechsimetka, line 47: maybe this could just be one component that returns different things depending whether tooltip is set? So:

```suggestion
export function Title({ label, tooltip }: TitleProps): ReactElement {
  const classes = useStyles()

  // There is no tooltip we can just return the label itself
  if (!tooltip) {
    return (
      <Typography className={classes.title} color="textSecondary" gutterBottom>
        {label}
      </Typography>
    )
  }

  // span is needed as Tooltip expects a non-functional element!
  return (
    <Tooltip title={tooltip}>
      <span>
        <RowBetween>
          <Typography className={classes.title} color="textSecondary" gutterBottom>
            {label}
          </Typography>
          <Info />
        </RowBetween>
      </span>
    </Tooltip>
  )
}
```

src/utils/threshold.ts
```diff
@@ -0,0 +1,95 @@
...
+  for (const item of thresholds) {
+    if (value >= item.minimumValue) {
+      return {
+        score: item.score,
+        maximumScore,
+        explanation: item.explanation,
+      }
+    }
+  }
+  throw Error(`Could not find any threshold for ${key} with value ${value}`)
```
vojtechsimetka, line 94: I'd rather return the default threshold with min value 0 than throwing error. I know this should never happen (the `value` is pretty much guaranteed to be `>0`), but you never know and since we are not catching the error, it would just break the whole app.

src/utils/threshold.ts
```diff
@@ -0,0 +1,95 @@
...
+  explanation: string
+}
+
+export type ThresholdValues = {
+  connectedPeers: ThresholdValue
+  population: ThresholdValue
+  depth: ThresholdValue
+}
+
+const GENERIC_ERROR = 'There may be issues with your Node or connection.'
```
agazso, line 29: I would not capitalize node. I would capitalize only if we were using `Bee` instead of `Node` because that is the name of the app. This way it may also give the impression that there are issues with Node.js.

vojtechsimetka (commented):
I usually prefer having just one component per file.

src/components/Row.tsx
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Elements within a row are vertically aligned to the center.
+ *
+ * Horizontal alignment is done by adding the maximum possible space between the elements.
+ */
+export function RowBetween({ children }: Props): ReactElement {
+  const classes = useStyles()
+
+  return <div className={classes.rowBetween}>{children}</div>
+}
```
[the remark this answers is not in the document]
Cafe137: Haha, perfect, switching to that. 🙂

vojtechsimetka (approved):

PR #172, opened 2021-08-18
feat: bee provider caching the state of the app and refreshing periodically

vojtechsimetka:
Simplified the loading system and error handling. The only page that is not affected is the accounting page which will be changed in a separate PR.

This makes it easy to provide the Bee and BeeDebug instances when the API url changes

https://user-images.githubusercontent.com/7974813/129854810-190d9cb2-4601-483c-8814-e01cadd1cd01.mov


Cafe137 (approved):
Top-notch refactor 🙂

PR #173, opened 2021-08-18
feat: changing API urls does not need the app refresh

### LINKED ISSUE #40: Bee and BeeDebug classes should be provided through context provider

### LINKED ISSUE #56: Changing bee urls should not require page reload
Bee and BeeDebug instances should be singletons living in some provider and all hooks should observe changes of the instance and refresh if the urls are changed.

vojtechsimetka:
Resolves #40 
Resolves #56

TODO:
- [x] Clear the interval and start new one with updated API values
- [x] Error handling when the Bee or the BeeDebug class is null (incorrect URL)
- [x] Refactor refresh function to `Promise.allSettled` so that error in the first API calls does not prevent subsequent API calls


src/containers/DepositModal.tsx
```diff
@@ -1,18 +1,24 @@
-import type { ReactElement } from 'react'
-import { beeDebugApi } from '../services/bee'
+import { ReactElement, useContext } from 'react'
+import { Context as SettingsContext } from '../providers/Settings'
 
 import WDModal from '../components/WDModal'
```
agazso, line 4: This seems to be used in both `DepositModal` and `WithdrawModal` so it seems to me that it would be better to call `WithdrawDepositModal` instead of `WDModal`

agazso (approved):

PR #177, opened 2021-08-20
chore: removed unnecessary and unused vars

vojtechsimetka:


Cafe137 (approved):
LGTM, I understand this is in a separate hook now.

vojtechsimetka:
Yes and no. It is in a separate hook, but that is only used in the provider. However, this object just does not need to be exposed from the provider at all.

PR #178, opened 2021-08-20
chore: renamed WDModal to WithdrawDepositModal

### LINKED ISSUE #175: Rename `WDModal` to `WithdrawDepositModal`
This seems to be used in both `DepositModal` and `WithdrawModal` so it seems to me that it would be better to call `WithdrawDepositModal` instead of `WDModal`

_Originally posted by @agazso in https://github.com/ethersphere/bee-dashboard/pull/173#discussion_r692796401_

vojtechsimetka:
resolves #175


Cafe137 (approved):
I would have also changed the import name `WDModal` to `WithdrawDepositModal`, but it probably makes sense to shorten it, so LGTM :)

vojtechsimetka:
Nice catch. I've changed the import names as well, makes sense!

PR #179, opened 2021-08-21
feat: remove the last update component

### LINKED ISSUE #176: Remove the "last updated" counter
After discussing with @agazso we agreed this does not provide much value for users and could confuse them

vojtechsimetka:
resolves #176


Cafe137 (approved):
LGTM :)

PR #181, opened 2021-08-23
fix: remove nested ternary operator and simplify ping peer functionality

### LINKED ISSUE #57: Eliminate nested ternary operators
Right now there is a number of nested ternary operators in several components. All of them have a `// FIXME` comment and should be removed.

vojtechsimetka:
Resolves #57

### DIFF
--- a/src/pages/peers/PeerTable.tsx
+++ b/src/pages/peers/PeerTable.tsx
@@ -27,28 +27,41 @@ interface Props {
   peers: Peer[] | null
 }
 
+interface PeerLatency {
+  rtt: string
+  loading: boolean
+}
+
+function getPingState(peerLatency: Record<string, PeerLatency>, peer: Peer): ReactElement {
+  if (peerLatency[peer.address]?.loading) return <CircularProgress size={20} />
+
+  if (peerLatency[peer.address]?.rtt) return <span>{peerLatency[peer.address]?.rtt}</span>
+
+  return <Autorenew />
+}
+
 function PeerTable(props: Props): ReactElement {
   const classes = useStyles()
   const { beeDebugApi } = useContext(SettingsContext)
 
-  const [peerLatency, setPeerLatency] = useState([{ peerId: '', rtt: '', loading: false }])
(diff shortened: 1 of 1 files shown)


src/pages/peers/PeerTable.tsx
```diff
@@ -27,28 +27,41 @@ interface Props {
...
     beeDebugApi
       ?.pingPeer(peerId)
       .then(res => {
-        setPeerLatency([...peerLatency, { peerId: peerId, rtt: res.rtt, loading: false }])
+        setPeerLatency({ ...peerLatency, [peerId]: { rtt: res.rtt, loading: false } })
       })
       .catch(() => {
-        setPeerLatency([...peerLatency, { peerId: peerId, rtt: 'error', loading: false }])
+        setPeerLatency({ ...peerLatency, [peerId]: { rtt: 'error', loading: false } })
       })
```
Cafe137, line 58: Hmm I am wondering if there is a race condition here. Not sure, but I'd probably use the arrow function for `setPeerLatency` which provides the previous value.

src/pages/peers/PeerTable.tsx
```diff
@@ -27,28 +27,41 @@ interface Props {
   peers: Peer[] | null
 }
 
+interface PeerLatency {
+  rtt: string
+  loading: boolean
+}
+
+function getPingState(peerLatency: Record<string, PeerLatency>, peer: Peer): ReactElement {
```
Cafe137, line 35: Hmm, first I found it a bit confusing (why is it not `PingState`?), but I understand this is being used in the `map` function, and that is why you chose this function style?

vojtechsimetka: Yeah exactly. I was considering if this should be a component on its own but that seemed quite bizarre. For optimalization purposes it's defined outside of the PeerTable component as this function does not need to be recreated with each re-render.

Cafe137 (approved):
Thanks, LGTM 🙂

PR #183, opened 2021-08-24
feat: separate info and status page

### LINKED ISSUE #168: Status text "Could not connect to Bee Node" is sometimes misleading
I have a running Node without peers.

Screenshot 2021-08-12 at 13 45 00

Since the peer check fails, it shows that I am not connected to my node.

Screenshot 2021-08-12 at 13 45 11

If any of the checks fail, we display this error message.

```
isOk={checks.every(c => c.isOk)}
```

It may be better to fine-tune it for some of the edge cases like this (by providing more accurate messages).

vojtechsimetka:
Resolves #168 (straight out removes this state)


src/pages/status/index.tsx
```diff
@@ -1,48 +1,170 @@
...
+    if (!isLoading) return
+
+    // Select first step that is not OK
+    // This is deliberately a for loop (and not forEach) so that we can terminate the useEffect from within the cycle
+    for (let i = 0; i < steps.length; ++i) {
+      if (!steps[i].isOk) {
+        setActiveStep(i)
+
+        return
+      }
```
Cafe137, line 109: Setting state in a loop - does it have any performance tax? Or is it irrelevant because it is in `useEffect`?

src/pages/status/index.tsx
```diff
@@ -1,48 +1,170 @@
...
+                disabled={isLoading}
+                onClick={() => setActiveStep(index === activeStep ? steps.length : index)}
+                style={{ justifyContent: 'space-between' }}
+              >
+                <div style={{ display: 'flex' }}>
+                  <div style={{ marginTop: '5px' }}>{label}</div>
+                  <div style={{ marginLeft: '12px' }}>
+                    {index === activeStep ? <ExpandLessSharp /> : <ExpandMoreSharp />}
+                  </div>
+                </div>
```
Cafe137, line 149: I think there are some refactoring opportunities here - like extracting some layout components, removing inline style, extracting logic. Not a blocker but I think this component has grown a bit too big 🙂

Cafe137 (approved):
See my question & refactoring ideas, otherwise I tested it and functions good and LGTM 🙂

src/pages/status/index.tsx
vojtechsimetka: I agree and I'll refactor it when I start implementing the new design (page by page).

PR #192, opened 2021-08-26
chore(deps): fix dependencies version

vojtechsimetka:


Cafe137 (approved):
Thanks for taking the time to hunt the dependency issue down 🙂

PR #194, opened 2021-08-26
style: sidebar styling

vojtechsimetka:
Please note that the sidebar is supposed to have a red circular background when there is an error and it is not selected. I was not able to style it that way and decided it would be better if for the time being I created an issue about #196 .

Screenshot 2021-08-31 at 10 01 36
Screenshot 2021-08-31 at 10 01 29
Screenshot 2021-08-31 at 10 01 01
Screenshot 2021-08-31 at 10 00 53


Cafe137 (approved):
LGTM :)

PR #204, opened 2021-09-09
feat: troubleshooting component

vojtechsimetka:
TODO:
- [x] rounded corners on the button component
- [ ] (will not tackle in this PR) ~icon on a button should be orange when not hovered but white when hovered over~
- [x] spacing in the text is slightly bigger than should be

Screenshot 2021-09-13 at 9 53 50


Cafe137 (approved):
LGTM :)

PR #240, opened 2021-11-11
feat: improve upload flow

Cafe137:
https://user-images.githubusercontent.com/77121044/143265328-e4617607-6338-447d-99c0-48ee1fd94e84.mp4


src/components/Container.tsx
```diff
@@ -0,0 +1,32 @@
...
+  textAlign?: 'left' | 'right' | 'center'
+}
+
+export default function Container(props: Props): ReactElement {
+  const classes = useStyles()
+
+  const inlineStyles: Record<string, string> = {}
+
+  props.maxHeight && (inlineStyles.maxHeight = props.maxHeight)
+  props.textAlign && (inlineStyles.textAlign = props.textAlign)
```
vojtechsimetka, line 25:
```suggestion
  if (props.maxHeight) inlineStyles.maxHeight = props.maxHeight
  if (props.textAlign) inlineStyles.textAlign = props.textAlign
```
I know that these are equivalent, but I'd prefer them as if statements.

src/components/ExpandableListItemActions.tsx
```diff
@@ -22,11 +22,13 @@ export default function ExpandableListItemActions({ children }: Props): ReactEle
   if (Array.isArray(children)) {
     return (
       <Grid container direction="row">
-        {children.map((a, i) => (
-          <Grid key={i} className={classes.action}>
-            {a}
-          </Grid>
-        ))}
+        {children
+          .filter(x => x)
```
vojtechsimetka, line 26: what does this do? This eliminates some undefined elements? (so `children` is array with holes?) Can you elaborate a bit how that can happen?

src/components/PaperGridContainer.tsx
```diff
@@ -0,0 +1,22 @@
...
+  },
+}))
+
+export function PaperGridContainer({ children }: Props): ReactElement {
+  const classes = useStyles()
+
+  return (
+    <Grid className={classes.grid} container justifyContent="space-between" alignItems="center" direction="row">
+      {children}
+    </Grid>
```
vojtechsimetka, line 20: I feel like this particular element is mixing layout with design. I would think about if it makes sense to separate those two. Although I do see why you did it this way.

src/components/SwarmButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  })
+
+  const classNames = className ? [className, classes.button].join(' ') : classes.button
+
+  return (
+    <Button className={classNames} onClick={() => onClick()} variant="contained" startIcon={icon}>
+      {children}
+    </Button>
+  )
+}
```
vojtechsimetka, line 46: Nice I like this! Would love to drop all the other buttons TBH.

src/components/VerticalSpacing.tsx
```diff
@@ -0,0 +1,11 @@
+// FIXME remove this lazy workaround
```
vojtechsimetka, line 1: please do :)

src/pages/files/FileIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import { ReactElement } from 'react'
+import { File } from 'react-feather'
+import { StripedWrapper } from '../../components/StripedWrapper'
+
+export function FileIcon(): ReactElement {
+  return (
+    <StripedWrapper>
+      <File />
```
vojtechsimetka, line 8: maybe this could be prop and then the `FileIcon.tsx` and `FolderIcon.tsx` could be merged?

src/pages/stamps/CreatePostageStampModal.tsx
```diff
@@ -138,7 +124,7 @@ export default function FormDialog({ label }: Props): ReactElement {
               <Field component={TextField} name="label" label="Label" fullWidth className={classes.field} />
             </DialogContent>
             <DialogActions>
-              <Button onClick={handleClose} variant="contained">
+              <Button onClick={() => onClose()} variant="contained">
```
vojtechsimetka, line 127:
```suggestion
              <Button onClick={onClose} variant="contained">
```

vojtechsimetka (commented):
Thanks a lot, looks quite good!

I was wondering, if you considered extracting the logic for file upload into a provider. It would clean up the component code and you could persist the state of the upload even if user navigates away. Plus if we are decide to, dragging and dropping files anywhere in the app could trigger the upload process.

vojtechsimetka:
Also if I try to download a folder that I just uploaded I get 404 error
Screenshot 2021-11-19 at 0 41 12

but it does work for both websites and single files

agazso (approved):

PR #244, opened 2021-11-15
feat: modularisation

nugaon:


package.json
```diff
@@ -36,17 +38,21 @@
...
     "qrcode.react": "1.0.1",
-    "react": "17.0.2",
     "react-copy-to-clipboard": "5.0.4",
-    "react-dom": "17.0.2",
     "react-feather": "2.0.9",
     "react-identicons": "1.2.5",
     "react-router-dom": "5.2.0",
     "react-syntax-highlighter": "15.4.4",
-    "semver": "7.3.5",
+    "semver": "^7.3.5",
```
vojtechsimetka, line 46: could we keep the dependencies locked?

src/config.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.BEE_DOCS_HOST = getProcessEnv('REACT_APP_BEE_DOCS_HOST') || 'https://docs.ethswarm.org/docs/'
+    this.BEE_DISCORD_HOST = getProcessEnv('REACT_APP_BEE_DISCORD_HOST') || 'https://discord.gg/eKr9XPv7'
+    this.GITHUB_REPO_URL =
+      getProcessEnv('REACT_APP_BEE_GITHUB_REPO_URL') || 'https://api.github.com/repos/ethersphere/bee'
+  }
+}
+
+export const config = new Config()
+
+export default config
```
vojtechsimetka, line 29: This can probably just be an object and not a full fledged class, right?

src/providers/Settings.tsx
```diff
@@ -42,6 +53,17 @@ export function Provider({ children }: Props): ReactElement {
     }
   }, [apiUrl])
 
+  useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log('bee-api-url2', beeApiUrl)
```
vojtechsimetka, line 58: please remove

webpack.config.ts
```diff
@@ -0,0 +1,67 @@
...
+          },
+        },
+      ],
+    },
+    // plugins: [
+    //   new ProvidePlugin({
+    //     React: 'react',
+    //     'window.React': 'react',
+    //   }),
+    // ],
```
vojtechsimetka, line 65:
```suggestion
    }
```

package.json
```diff
@@ -36,17 +38,21 @@
     "notistack": "1.0.10",
     "opener": "1.5.2",
     "qrcode.react": "1.0.1",
-    "react": "17.0.2",
```
vojtechsimetka, line 39: this should stay as dependency

package.json
```diff
@@ -36,17 +38,21 @@
...
     "react-copy-to-clipboard": "5.0.4",
-    "react-dom": "17.0.2",
```
vojtechsimetka, line 41: this should stay as dependency

vojtechsimetka (commented):
Two general comments:
- Please lock the versions
- Do we need `babel` and `webpack`? My fear is that this will start clashing (version wise) with `react-scripts` (got that experience in past) and this pattern is heavily discouraged.

BTW: Great job cleaning up the config :)

src/config.ts
nugaon: can be. I made it like this because of the `readonly` prop.
what would be the advantage of that?

vojtechsimetka: just better readability IMHO but let's keep it as is, you are right it's less error prone this way.

vojtechsimetka (commented):
Looks good, please do lock all the dependencies if possible (especially as this is now going to be included in other projects).

nugaon:
> Looks good, please do lock all the dependencies if possible (especially as this is now going to be included in other projects).

as we discussed on one of our roundtable, it is not really necessary if you use `npm ci` instead of `npm i`.

p.s. since then some wild conflicts appeared I'm going to make a force push after resolving those.

vojtechsimetka:
> > Looks good, please do lock all the dependencies if possible (especially as this is now going to be included in other projects).
> 
> as we discussed on one of our roundtable, it is not really necessary if you use `npm ci` instead of `npm i`.

Can you please refer me to that conclusion/discussion? Maybe it was done in my absence. AFAIK I presented this topic on roundtable in august and there the conclusion was that at least for the UI project we use fixed version (although this is not recorded in the notes). I am happy to reopen this discussion, just wanted to follow our guidelines - https://roamresearch.com/#/app/swarm-org/page/K97knoZCm

nugaon:
> AFAIK I presented this topic on roundtable in august and there the conclusion was that at least for the UI project we use fixed version (although this is not recorded in the notes)

as I remember it was after that when I presented in npm security topic.

Nevertheless, I upload the package.json with the fixed version numbers, because it cannot cause any bad thing it is just incovenient to do so. That's why `package-lock.json` exists.

vojtechsimetka (approved):
LGTM. Will probably have to rebase/resolve conflicts with master

nugaon:
> LGTM. Will probably have to rebase/resolve conflicts with master

there are no conflicts.

PR #246, opened 2021-11-16
feat: add dev mode flag

Cafe137:
Usage:

`REACT_APP_DEV_MODE=1 npm start`

### DIFF
--- a/src/providers/Bee.tsx
+++ b/src/providers/Bee.tsx
@@ -1,19 +1,18 @@
-import type { ChequebookBalance, Balance, Settlements } from '../types'
-import { createContext, ReactChild, ReactElement, useEffect, useState, useContext } from 'react'
-import { Token } from '../models/Token'
-import semver from 'semver'
-import { engines } from '../../package.json'
-import { Context as SettingsContext } from './Settings'
-
 import type {
-  NodeAddresses,
   ChequebookAddressResponse,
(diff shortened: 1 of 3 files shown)


agazso:
If you add the same condition to the `chequebook` as well then it will work with Bee in dev mode.

The only hickup I saw is that the overlay address on the Info page will show `undefined[...]undefined`, see the attached screenshot:
Screenshot 2021-11-16 at 18 19 20

Cafe137:
@agazso Can you please check?

agazso:
> @agazso Can you please check?

works well

src/providers/Bee.tsx
```diff
@@ -113,11 +112,13 @@ function getStatus(
     blockchainConnection: Boolean(nodeAddresses?.ethereum),
     debugApiConnection: Boolean(debugApiHealth?.status === 'ok'),
     apiConnection: apiHealth,
-    topology: Boolean(topology?.connected && topology?.connected > 0),
+    // FIXME: `REACT_APP_DEV_MODE` is a temporary workaround to be able to develop with only one node
```
vojtechsimetka, line 115: Would you mind adding this to the readme? That also increases the chance we don't forget to remove it :)

vojtechsimetka (approved):
Thanks, works well!

PR #259, opened 2021-11-29
ci: upload build to testnet

AuHau:


.github/workflows/check.yaml
```diff
@@ -68,3 +68,8 @@ jobs:
         uses: ethersphere/beeload-action@v1
         with:
           preview: 'true'
+
+      - name: Upload to testnet
+        uses: ethersphere/beeload-action@v1
+        with:
+          bee-url: https://api.gateway.testnet.ethswarm.org
```
vojtechsimetka, line 75: Should this be uploaded to both mainnet and testnet?

vojtechsimetka: It will probably overwrite the comment and only the one to testnet shown

AuHau: This upload does not generate the "PR Preview" comment so nothing is overridden.  The idea here is only to generate traffic on the testnet. Originally I wanted to make full switch, but there is currently no `bzz.link` on testnet and in the end we don't want to deploy it there just yet.

Cafe137 (approved):

PR #260, opened 2021-11-29
feat: add website and folder upload and download

Cafe137:
image

image


package.json
```diff
@@ -50,6 +53,7 @@
     "@commitlint/config-conventional": "14.1.0",
```
vojtechsimetka, line 53: somewhere we lost `react-router` dependency. Please add it `npm i react-router --save`. Then check the depcheck with `npm run depcheck`. We however do have the `@types/react-router`.

src/pages/files/Share.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { AssetSummary } from './AssetSummary'
+import { DownloadActionBar } from './DownloadActionBar'
+
+interface Props {
+  match: {
+    params: {
+      hash: string
+    }
+  }
+}
```
vojtechsimetka, line 24: There is a shorthand for this

```ts
import { RouteComponentProps } from 'react-router';

// example route
<Route path="/products/:name" component={ProductContainer} />

interface MatchParams {
    name: string;
}

interface Props extends RouteComponentProps<MatchParams> {
}
```

src/pages/files/Share.tsx
```diff
@@ -0,0 +1,124 @@
...
+      setFiles(convertManifestToFiles(entries))
+    }
+  }
+
+  function onOpen() {
+    window.open(`${apiUrl}/bzz/${reference}/`, '_blank')
+  }
+
+  function onClose() {
+    if (history.length < 2) {
```
vojtechsimetka, line 66: a bit surprised by this number `2` here. I'd suggest some comment why 2 :)

src/utils/hash.ts
```diff
@@ -0,0 +1,3 @@
+export function shortenHash(hash: string, sliceLength = 8): string {
+  return hash.slice(0, sliceLength) + '[…]' + hash.slice(-sliceLength)
```
vojtechsimetka, line 2: please use template instead of string "math"

src/utils/local-storage.ts
```diff
@@ -0,0 +1,70 @@
...
+    createdAt: Date.now(),
+    hash,
+    name,
+  })
+
+  if (history.length > 10) {
+    history.length = 10
+  }
+  localStorage.setItem(key, JSON.stringify(history))
+}
```
vojtechsimetka, line 33: I'd suggest to just always serialize the history you have live in app and overwrite the localstorage history key. I don't see a reason why this should always try retrieve the full history, deserialize it, find a record that may need to be replaced, serialize it and save it.

vojtechsimetka (commented):
Good job! I like this a lot and left just a few minor comments. I am still testing it but once done I'll approve the PR.

I've resolved the FOSSA issues (it sometimes can not handle well dependencies that are licensed under two different licenses - in this case MIT & GPL3.0.

vojtechsimetka (approved):
Awesome job, thanks! :)

PR #267, opened 2021-12-13
fix(build): lib

nugaon:


.github/workflows/check.yaml
```diff
@@ -66,6 +66,9 @@ jobs:
 
       - name: Build
         run: npm run build
+      
+      - name: Build Component
+        run: npm build:component
```
vojtechsimetka, line 71: I would also consider type compilation `compile:types`, or do you think it is not needed as we already do `check:types` above?

vojtechsimetka (approved):

.github/workflows/check.yaml
nugaon: yeah, it should be enough

vojtechsimetka, line 71:
```suggestion
        run: npm run build:component
```

PR #270, opened 2021-12-15
fix: package.json files

nugaon:
The types and the assets are not available in the npmjs package of the library at this moment.

This PR will fix that for the next build and release.

Thanks @AuHau for the tip!

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "check:types": "tsc --project tsconfig.lib.json"
   },
   "files": [
+    "lib",
     "build",
     "serve.js"
   ],


.github/workflows/publish_npmjs.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: npm run compile:types
-      - run: npm build:component
+      - run: npm run build:component
```
vojtechsimetka, line 19: I am a bit confused, this change is already in the master branch, is this some github bug or are you somehow merging against some older commit (that looks like master?) 

https://github.com/ethersphere/bee-dashboard/blob/master/.github/workflows/publish_npmjs.yaml

vojtechsimetka (commented):
@nugaon did you try and publish this in some fork of this repo? I want to make sure that this fixes all the problems you have to import bee-dashboard in the swarm-extension.

Cafe137 (approved):

.github/workflows/publish_npmjs.yaml
nugaon: hmmm, it is strange indeed

nugaon: I fixed. I think it was based on an older commit

PR #280, opened 2022-01-09
chore(deps): update react router from v5 to v6

vojtechsimetka:
I was doing this on my personal project so I decided to utilise my migration knowledge here :).


src/pages/feeds/UpdateFeed.tsx
```diff
@@ -76,10 +73,10 @@ export default function UpdateFeed(props: RouteComponentProps<MatchParams>): Rea
     }
 
     try {
-      await updateFeed(beeApi, identity, props.match.params.hash, selectedStamp, password as string)
+      await updateFeed(beeApi, identity, hash!, selectedStamp, password as string)
```
vojtechsimetka, line 76: I don't really know under what circumstances the hash would be undefined. Anyways `updateFeed` function does validation AFAIK.

src/pages/files/Share.tsx
```diff
@@ -19,17 +19,14 @@ import { AssetPreview } from './AssetPreview'
...
-}
-
-export function Share(props: RouteComponentProps<MatchParams>): ReactElement {
+export function Share(): ReactElement {
   const { apiUrl, beeApi } = useContext(SettingsContext)
   const { status } = useContext(BeeContext)
 
-  const reference = props.match.params.hash
+  const { hash } = useParams()
+  const reference = hash!
```
vojtechsimetka, line 27: I don't think hash can be undefined, but we should consider if it should be validated here.

src/pages/files/Share.tsx
```diff
@@ -72,15 +69,15 @@ export function Share(props: RouteComponentProps<MatchParams>): ReactElement {
 
   function onClose() {
     // POP means there is no history - nowhere to go back yet
```
Cafe137, line 71: Just a minor thing, can you please remove this comment? It is a bit confusing now 🙂

src/pages/files/Share.tsx
```diff
@@ -72,15 +69,15 @@ export function Share(props: RouteComponentProps<MatchParams>): ReactElement {
...
-    if (history.action === 'POP') {
```
Cafe137, line 75: This seems to be the other way around

Cafe137 (changes requested):
Thanks for the upgrade 🙂

Cafe137 (approved):

PR #285, opened 2022-01-19
refactor: remove google fonts dependency

Cafe137:


src/App.css
```diff
@@ -1,34 +1,148 @@
...
+/* latin */
+@font-face {
+  font-family: 'Work Sans';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: url(assets/fonts/WorkSans/WorkSans-Regular.ttf) format('truetype');
+  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC,
+    U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+}
```
vojtechsimetka, line 56: I am not too familiar with font types, but do we need `vietnamese` and `latin-ext`? I don't think we use those. Maybe we don't need to specify the unicode-range and just import the whole font?

vojtechsimetka (commented):
Do we use all these font weights? 300, 400, 500 & 600?

PR #286, opened 2022-01-19
fix: get current price from chain state

Cafe137:


src/utils/index.ts
```diff
@@ -186,6 +187,11 @@ export function convertAmountToSeconds(amount: number): number {
   return amount / 10 / 1
 }
 
-export function calculateStampPrice(depth: number, amount: number): number {
-  return (amount * 2 ** (depth - 16) * 2) / 1e16
+export function calculateStampPrice(depth: number, amount: number, currentPrice?: NumberString): number {
```
vojtechsimetka, line 190: I dont think the `currentPrice` parameter should be optional. If we don't have `currentPrice` we should not estimate it.

vojtechsimetka (approved):

Issue #243, opened 2021-11-12
Support routing on single page apps


vojtechsimetka:
When uploading websites set the `swarm-error-document` to `index.html`. Later on this could be configurable.

PR #287, opened 2022-01-19
feat: add hash based routing

Cafe137:


vojtechsimetka (approved):
I am ok with this, but I want to understand the motivation. Is it to solve #243 ?

PR #311, opened 2022-03-23
feat: add bee desktop toolkit

Cafe137:


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzafkpycdkf5ju7znbkxnaquecoaz6ch4ik65zzmfqqvjl37l5elrda.bzz.link
**Swarm Hash:** `2a9f810d45ea69fcb42abb410a104e067c23f10af773961610aa57bfafa45c46`
**Commit Hash:** c1a5ec555dd5244711fba76c5edb5e28c3bf4f7a
**Commit Message:** feat: add light node upgrade

vojtechsimetka:
@Cafe137 what is your idea about this? Should we be rebasing this on master continuously (maybe after we shave off some of the already implemented functionality)? Or do you want to merge this?

src/layout/Dashboard.tsx
```diff
@@ -33,7 +29,6 @@ const Dashboard = (props: Props): ReactElement => {
       <Container className={classes.content}>
         <ErrorBoundary>
           <>
-            <AlertVersion />
```
Cafe137, line 36: What should we do about AlertVersion?

src/utils/rpc.ts
```diff
@@ -0,0 +1,65 @@
+import { Contract, providers } from 'ethers'
+
+const PROVIDER = 'https://gno.getblock.io/mainnet/?api_key=d7b92d96-9784-49a8-a800-b3edd1647fc7'
```
Cafe137, line 3: I'm OK with this until we have FDS RPC working from browser

src/layout/Dashboard.tsx
vojtechsimetka: IMHO it should stay in general but we should make it less strict.

src/providers/Bee.tsx
```diff
@@ -61,6 +61,14 @@ interface ContextInterface {
...
 }
 
+const urlSearchParams = new URLSearchParams(window.location.search)
+const newApiKey = urlSearchParams.get('v')
+
+if (newApiKey) {
+  localStorage.setItem('apiKey', newApiKey)
+  window.location.search = ''
+}
+
```
vojtechsimetka, line 71: Could you extract this into `SettingsProvider` and store there? That is the top level provider and IMHO makes sense to also store the key in info. It already passes down the `Bee` object.

src/utils/desktop.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function getBeeEthereumAddress(): Promise<string> {
+  const status = await getDesktopStatus()
+
+  if (!status.address) {
+    throw Error('Bee node Ethereum address is not yet available')
+  }
+
+  return status.address
+}
```
vojtechsimetka, line 24: This is a duplication, the Bee provider already has the ethereum address. I am fine with it, just wanted to mention this.

src/utils/net.ts
```diff
@@ -0,0 +1,33 @@
...
+  const authorization = localStorage.getItem('apiKey')
+
+  if (!authorization) {
+    throw Error('API key not found in local storage')
+  }
+  const headers = {
+    'content-type': 'application/json',
+    authorization,
+  }
+  const response = await fetch(url, {
```
vojtechsimetka, line 25: can we just use `axios`? We use it elsewhere.

src/utils/rpc.ts
```diff
@@ -0,0 +1,65 @@
...
+
+async function eth_getBalance(address: string): Promise<string> {
+  if (!address.startsWith('0x')) {
+    address = `0x${address}`
+  }
+  const response = await fetch(PROVIDER, {
```
vojtechsimetka, line 9: can you use axios?

src/providers/Bee.tsx
```diff
@@ -209,6 +222,22 @@ export function Provider({ children }: Props): ReactElement {
...
+      if (xdai?.then) {
+        xdai.then(balance => setXdai(balance))
+      }
+
+      if (bzz?.then) {
+        bzz.then(balance => setBzz(balance))
+      }
+    }
+  }, [nodeAddresses])
+
```
Cafe137, line 240: @vojtechsimetka It seemed a good idea to put the RPC balance checks in a `useEffect` hook depending on `nodeAddresses`. However that triggered many duplicate requests at the same time, so I had to introduce debouncing. Lmk what you think of this approach.

src/utils/desktop.ts
Cafe137: Removed it, no longer needed

src/providers/Bee.tsx
vojtechsimetka: Yeah this is good. But I think even putting `[nodeAddresses?.ethereum]` in the hook could solve it (so at the beginning it is null and then always the same address. But I am also ok with what you implemented.

src/providers/Bee.tsx
```diff
@@ -7,16 +8,21 @@ import {
...
 import { useLatestBeeRelease } from '../hooks/apiHooks'
 import { Token } from '../models/Token'
 import type { Balance, ChequebookBalance, Settlements } from '../types'
+import { Rpc } from '../utils/rpc'
 import { Context as SettingsContext } from './Settings'
 
+interface RpcBalance {
+  bzz: string
+  xdai: string
+}
```
vojtechsimetka, line 24: Can these be instances of Token already?

src/utils/desktop.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function getDesktopStatus(): Promise<DesktopStatus> {
+  const response = await getJson(`http://${getDesktopHost()}/status`)
+
+  return response as DesktopStatus
+}
+
+export async function getGasFromFaucet(address: string): Promise<void> {
+  await fetch(`http://getxdai.co/${address}/0.1`, {
```
vojtechsimetka, line 17: this still uses fetch

vojtechsimetka (approved):
Thanks, this looks very good :)

PR #352, opened 2022-04-28
feat: wait for postage stamp to be usable when bying it

### LINKED ISSUE #321: Buying postage stamp does not wait for it to be usable
This is probably not a problem on the stamps management page, but a real issue when trying to upload new file.

vojtechsimetka:
resolves #321 

This is a very minimal implementation of the functionality. I would like to point out that the UX here is pretty bad as it does not tell the user what is happening (e.g. the stamp was bought waiting for it to be usable) and if for whatever reason the stamp is not usable until the `DEFAULT_STAMP_USABLE_TIMEOUT = 120_000` is reached, it will fail with error message. My suggestion would be to merge this PR but have a design session with David how to improve the UX. On mainnet purchasing the stamp will take 30-90 seconds and then waiting for it to be usable also takes up to 90 seconds.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -2,6 +2,7 @@ import { BigNumber } from 'bignumber.js'
 import { Token } from '../models/Token'
 import { decodeCid } from '@ethersphere/swarm-cid'
 import { BZZ_LINK_DOMAIN } from '../constants'
+import { BatchId, BeeDebug, PostageBatch } from '@ethersphere/bee-js'
 
 /**
  * Test if value is an integer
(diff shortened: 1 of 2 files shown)


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgza5dkxye7pzanzvpyubrm3hvgoumhrdn4ulo4m2m74c4qhl7y7rksq.bzz.link
**Swarm Hash:** `e8d57c13efc81b9abf140c59b3d4cea30f11b7945bb8cd33fc172075ff1f8aa5`
**Commit Hash:** 58b2afc063ae70b10eae8cd2797e96f52a59df51
**Commit Message:** feat: wait for postage stamp to be usable when bying it

src/utils/index.ts
```diff
@@ -216,3 +217,37 @@ export function shortenText(text: string, length = 20, separator = '[…]'): str
...
+    )
+
+  const stampWaitPromise = async (): Promise<PostageBatch | undefined> => {
+    while (!timeoutReached) {
+      const stamp = await beeDebug.getPostageBatch(batchId)
+
+      if (stamp.usable) return stamp
+      await sleepMs(pollingFrequency)
+    }
+  }
```
Cafe137, line 249: I usually do this with for loops, I think the upside is you don't need to race promises, no need for a separate setTimeout for the rejection, and the implementation is a single block that is simpler IMHO.

I am fine with this as well, just wondering if this has some pros I am unaware of 🙂

vojtechsimetka: Hey! Actually no benefits, mostly really just complications. I will refactor, thanks for the suggestion :)

vojtechsimetka: The only advantage is that it spends exactly the `timeout` time unlike for loop which also adds the time spent on requests to check the stamp is usable . I don't see that as a problem, just wanted to point it out :)

Cafe137 (approved):

PR #354, opened 2022-04-29
fix: remove restrictions on postage stamp label

### LINKED ISSUE #326: Cannot buy stamp with space in the label
I cannot buy a stamp if there is space or other special characters (like dash) in the label field, the button becomes disabled. I am not sure where this limitation comes from. 

If there is a limitation for label names in Bee then it would be good to display a warning with an explanation of what is an allowed label name. Or alternatively we could only allow those characters in the field which make up valid label names.

vojtechsimetka:
resolves #326

I've tested this and it works fine with any UTF-8 string on bee dev and bee-factory (even put it some evil czech characters :) )
Screenshot 2022-04-29 at 10 54 06

### DIFF
--- a/src/pages/stamps/PostageStampCreation.tsx
+++ b/src/pages/stamps/PostageStampCreation.tsx
@@ -113,9 +113,6 @@ export function PostageStampCreation({ onFinished }: Props): ReactElement {
           else if (amount.isLessThanOrEqualTo(0)) errors.amount = 'Amount must be greater than 0'
         }
 
-        // Label
-        if (values.label && !/^[0-9a-z]*$/i.test(values.label)) errors.label = 'Label must be an alphanumeric string'
-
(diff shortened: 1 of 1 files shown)


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzayulhra6n5bcgcgl5gt6qz2czf34kyq3uoi2x4cx2rsesvupebm4a.bzz.link
**Swarm Hash:** `c5167883cde84461197d34fd0ce8592ef8ac437472357e0afa8c892ad1e40b38`
**Commit Hash:** 7bdb6fde66ddc79c4a55c984f35b33cd9b575b25
**Commit Message:** fix: remove restrictions on postage stamp label

Cafe137 (approved):
Nice :)

PR #365, opened 2022-05-11
fix: nested directory upload preserves the directory structure

vojtechsimetka:
resolves #363

PR #372, opened 2022-05-26
feat: gift wallet

Cafe137:


src/components/CashoutModal.tsx
```diff
@@ -61,7 +61,7 @@ export default function CheckoutModal({ peerId, uncashedAmount }: Props): ReactE
   return (
     <div>
       <Button variant="contained" onClick={handleClickOpen} startIcon={<Zap size="1rem" />}>
-        Cash out peer {peerId.substr(0, 8)}[…]
+        Cash out peer {peerId.slice(0, 8)}[…]
```
vojtechsimetka, line 64: Is there a reason why `slice` is better than `substr`?

src/models/Token.ts
```diff
@@ -78,4 +80,11 @@ export class Token {
 
     return asString.slice(0, indexOfSignificantDigit + digits)
   }
+
+  minusEther(amount: string): Token {
```
vojtechsimetka, line 84:
```suggestion
  minusBaseUnits(amount: string): Token {
```
Also does it make sense to have `amount` as `bigint` and `BigNumber` in addition to string.

src/pages/files/Upload.tsx
```diff
@@ -83,9 +83,9 @@ export function Upload(): ReactElement {
         // The website is in some directory, remove it
         if (idx.commonPrefix) {
           const substrStart = idx.commonPrefix.length
-          indexDocument = idx.indexPath.substr(substrStart)
-          fls = files.map(f => {
-            const path = (f.path as string).substr(substrStart)
+          indexDocument = idx.indexPath.slice(substrStart)
+          fls = fls.map(f => {
```
vojtechsimetka, line 87:
```suggestion
          fls = files.map(f => {
```
This would reintroduce bug #365

vojtechsimetka (changes requested):

src/components/CashoutModal.tsx
vojtechsimetka: Oh I guess it is because `substr` is being deprecated?

src/layout/Dashboard.tsx
```diff
@@ -33,7 +29,6 @@ const Dashboard = (props: Props): ReactElement => {
       <Container className={classes.content}>
         <ErrorBoundary>
           <>
-            <AlertVersion />
```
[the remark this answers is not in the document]
agazso: I asked for removing this for the internal testing. It would be good to improve this, because the current version is very annoying (it comes back after a reload or when you open the page from the menu) and it requires us to have exactly the same version of Bee bundled with the desktop as the version that the bee-js supports that is bundled with the dashboard.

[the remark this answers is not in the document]
vojtechsimetka: OK we can remove it in this PR, but we can definitely reintroduce it with better mechanism. `bee-js`has a different method to check version as that is less strict. And we can also remember the dismissal of the warning with specific version and timestamp to show it when the version changes or when e.g. 7 days pass.

bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzaqxulgxrhkxq3do766qkqaenl26dtwbomgamnxh4qs4mmewyntj3q.bzz.link
**Swarm Hash:** `85e8b35e2755e1b1bbfef4150011abd7873b05cc3018db9f909718c25b0d9a77`
**Commit Hash:** 25c11581b631f50d780c62978344ab243b09642c
**Commit Message:** build: remove unused dependency

vojtechsimetka (approved):
LGTM and works fine

PR #385, opened 2022-06-16
feat: sentry integration

AuHau:
This PR tackles several things:

 - Fixes the Env. variable usage for configs
 - Creates workflow and support to develop Dashboard against the Bee Desktop API 
 - Enhance CI to notify Sentry about new releases and upload Source maps to it
 - If Sentry Key is defined at build time, then error tracking and error boundaries are used and enabled
 - Creates `ItsBroken` component that is displayed on errors (TODO: Some nicer design would be nice)

Related to https://github.com/ethersphere/bee-desktop/pull/105


Cafe137 (approved):

.github/workflows/publish_npmjs.yaml
```diff
@@ -20,3 +20,11 @@ jobs:
...
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+      - name: Create Sentry release
+        uses: getsentry/action-release@v1
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+        with:
+          sourcemaps: ./build/static/js
```
vojtechsimetka, line 30: How does this work? The sentry secrets are stored in the build files? So technically this leaks, right?

package.json
```diff
@@ -32,12 +32,14 @@
...
     "@material-ui/lab": "4.0.0-alpha.57",
+    "@sentry/react": "^7.1.1",
+    "@sentry/tracing": "^7.1.1",
     "assert": "^2.0.0",
     "axios": "0.24.0",
     "bignumber.js": "9.0.1",
     "buffer": "^6.0.3",
-    "crypto-browserify": "^3.12.0",
     "crypto": "npm:crypto-browserify",
+    "crypto-browserify": "^3.12.0",
```
vojtechsimetka, line 42: do we need both?

src/App.tsx
```diff
@@ -14,42 +16,88 @@ import { Provider as StampsProvider } from './providers/Stamps'
...
+  if (config.SENTRY_KEY) {
+    return (
+      <Sentry.ErrorBoundary
+        showDialog
+        fallback={({ error, componentStack, resetError }) => <ItsBroken message={error.message} />}
+      >
+        {mainApp}
+      </Sentry.ErrorBoundary>
+    )
+  }
```
vojtechsimetka, line 98: not that we are relying on it in CSS, but please note that this changes the component nesting. Before we had directly `<div id="app">...</div>` now this is nested in the sentry error boundary.

vojtechsimetka (approved):
LGTM

package.json
AuHau: Yes. The `crypto` is only an alias for the later one.

PR #388, opened 2022-06-16
feat: sentry feedback form

AuHau:
This PR adds feedback form. I am not sure about the styling and design though so this is just an initial version and I am looking for feedback.


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzais34wtoeygxuxg47vagztzo7yfqi3pyi7wx4zyh5lper56zujr6a.bzz.link
**Swarm Hash:** `44b7cb4dc4c1af4b9b9fa80d99e5dfc1608dbf08fdafcce0fd5bc91efb344c7c`
**Commit Hash:** 47c6a2cc080f50786c8a82cf22fea3835d73dde6
**Commit Message:** feat: sentry feedback form

david-swarm:
I will propose a simple design improvement today

AuHau:
@david-swarm be aware that the only thing that is easy to modify is the actual trigger link (the "Give us feedback!" at the bottom). The form itself is provided by the third-party and we can only tweak texts (see [here](https://docs.sentry.io/platforms/javascript/enriching-events/user-feedback/#customizing-the-widget)) and not much more. It is possible to create our own implementation, but for time constrain I would suggest to not doing that at this moment.

david-swarm:
@AuHau understood. Just out of curiosity, simply changing fonts without changing any of the window layout?

vojtechsimetka (approved):

AuHau:
@david-swarm I will try it, but honestly not sure if by modifying it like this, we are not breaching some license conditions.

AuHau:
@david-swarm so I have updated the fonts and some wordings. Let me know WDYT. Also is the orange underline for hover over the "Give us feedback!" alright?

david-swarm:
Screenshot 2022-06-17 at 15 45 55

Design wise it would be awesome to treat it consistently with the other sidebar items. @vojtechsimetka I checked and in the latest design even the node status is consistent with the other items :)

AuHau:
>  Let's find a better placeholder for this. Even something like "Please let us know your thoughts..."

Yeah, would love to, but I am afraid, not possible without our own implementation :-( 

I will try to unify the Send feedback button.

agazso:
> > Let's find a better placeholder for this. Even something like "Please let us know your thoughts..."
> 
> Yeah, would love to, but I am afraid, not possible without our own implementation :-(
> 
> I will try to unify the Send feedback button.

@david-swarm could you please design a feedback popup with a textbox and a button? It seems like it would be easier to build a custom one than trying to work around the builtin one from Sentry.

AuHau:
@agazso is that a priority though now?

agazso:
> @agazso is that a priority now though? IMHO it is easier to go with what we have now as IMHO it is sufficient. Most users won't even see this. And the thing they see (eq. the Send feedback button) will be polished and unified.

Implementation always follows design, so the sooner we have the design the sooner we can have it implemented. Most likely not next week, but eventually it would be good to have something that is inline with the rest of the design.

vojtechsimetka:
So @AuHau can we merge?

AuHau:
No

AuHau:
@vojtechsimetka can you re-review please and then IMHO it is ready to merge.

vojtechsimetka (approved):

PR #438, opened 2022-06-24
feat(ultra-light): block features that are not available in this mode

vojtechsimetka:
Account page only shows wallet setup
Screenshot 2022-06-24 at 12 25 32

File page only shows download
Screenshot 2022-06-24 at 12 35 10


agazso (approved):

AuHau (approved):
LGTM

PR #456, opened 2022-06-29
feat: transfer everything out of the gift wallet

vojtechsimetka:
So this is by no means perfect:

- It does not have any retry mechanism
- it does not correctly check on the UI if the balance is enough for both of the transfers.

But I think it is improvement over what is there right now.  I would like to do a complete refactor of this code later (we are creating providers all the time, I think we can better encapsulate things etc).

### DIFF
--- a/src/utils/rpc.ts
+++ b/src/utils/rpc.ts
@@ -1,6 +1,6 @@
 import { debounce } from '@material-ui/core'
 import { Contract, providers, Wallet, BigNumber as BN } from 'ethers'
-import { bzzABI } from './bzz-abi'
+import { bzzABI, BZZ_TOKEN_ADDRESS } from './bzz-abi'
 
 async function eth_getBalance(address: string, provider: providers.JsonRpcProvider): Promise<string> {
   if (!address.startsWith('0x')) {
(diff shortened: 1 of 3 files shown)


src/utils/rpc.ts
```diff
@@ -79,7 +70,7 @@ export async function sendBzzTransaction(
 ): Promise<TransferResponse> {
   const signer = await makeReadySigner(privateKey, jsonRpcProvider)
   const gasPrice = await signer.getGasPrice()
-  const bzz = new Contract('0xdBF3Ea6F5beE45c02255B2c26a16F300502F68da', bzzContractInterface, signer)
+  const bzz = new Contract('0xdBF3Ea6F5beE45c02255B2c26a16F300502F68da', bzzABI, signer)
```
agazso, line 73: It would be good to have the contract address as a const.

bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzatvt7bbjryckc2ugdzugxxzix6hl32yfqsaem62z6kbcwlfb3cgwq.bzz.link
**Swarm Hash:** `9d67f08531c0942d50c3cd0d7be517f1d7bd60b09008cf6b3e504565943b11ad`
**Commit Hash:** 507aa5ff23c2cff6057ead31f6cc772bf4d20118
**Commit Message:** feat: transfer everything out of the gift wallet

agazso (approved):
Looks good. Let's try to add retry logic later.

PR #489, opened 2022-07-22
feat: log errors to console when showing error notification

### LINKED ISSUE #473: Whenever there is an error notification it would be good to log to console
Currently there are error notifications but they don't always tell much (and sometimes they tell too much, see #457). However they are not printed on the console, so it is almost impossible to debug what the problem is.

It would be good to establish a rule that anytime there is an error notification, there is an accompanied `console.error` call so that the error can be investigated.

Example:

Screenshot 2022-07-05 at 18 09 34

vojtechsimetka:
Resolves #473

unfortunately, there is no hook that could be used in the `notistack`

### DIFF
--- a/src/components/CashoutModal.tsx
+++ b/src/components/CashoutModal.tsx
@@ -48,6 +48,7 @@ export default function CheckoutModal({ peerId, uncashedAmount }: Props): ReactE
           )
         })
         .catch((e: Error) => {
+          console.error(e) // eslint-disable-line
           enqueueSnackbar(<span>Error: {e.message}</span>, { variant: 'error' })
         })
         .finally(() => {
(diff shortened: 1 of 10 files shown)


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgza5nqmx6spdq4xw6kv23k45rtahqad6m5j6t5vqfiv3giqxdtovlhq.bzz.link
**Swarm Hash:** `eb60cbfa4f1c397b7955d6d5cec6603c003f33a9f4fb581515d9910b8e6eaacf`
**Commit Hash:** a7bf36c2fe8fec654627684e066660f5a93c06fb
**Commit Message:** feat: log errorss to consolve when showing error notification

Cafe137 (approved):

PR #513, opened 2022-08-02
build: bundle assets inline for the component library

vojtechsimetka:


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzaomrfntsoyxnsfhrhuezypdpfqgcr54v2tkmwdnfee7eiwwzuaxja.bzz.link
**Swarm Hash:** `732256ce4ec5db229e27a133878de581851ef2ba9a9961b4a427c88b5b3405d2`
**Commit Hash:** 723c9c958a92d94ca089cd9b573dd37892ee9ba4
**Commit Message:** chore(deps): allow react version 17.x.x and 18.x.x

package.json
```diff
@@ -50,9 +50,9 @@
     "notistack": "1.0.10",
     "opener": "1.5.2",
     "qrcode.react": "1.0.1",
-    "react": ">= 17.0.2",
+    "react": ">=17.0.0 || >=18.0.0",
```
AuHau, line 53: Why the `||`? I mean `18` is greater then `17`...

AuHau (approved):
I don't understand the `||` but otherwise LGTM! Pretty dope work! 🎉 Thanks!

package.json
vojtechsimetka: yes but `>=` operator does not work for different major version so with `>=17.0.0` it only includes versions starting with `17.x.x`. React version `18.x.x` would not be considered.

Issue #518, opened 2022-08-05
Swarm Extension: process is undefined


vojtechsimetka:
When using the bee-dashboard component in Swarm Extension, it does not render because Chrome does not have process.

**Options:**
1. Create a util function that returns env variable or undefined `getEnvVar = (varName) => process?.env?[varName]`
2. Alter webpack component build to substitute at build time `REACT_APP_...` variables with values
3. Completely remove env variables and let everything be configurable through component props

![Screenshot 2022-08-04 at 22 06 33](https://user-images.githubusercontent.com/7974813/183006545-f7b4e0a1-cac6-486a-bd51-e5a55df7a827.png)
![Screenshot 2022-08-04 at 22 06 42](https://user-images.githubusercontent.com/7974813/183006548-bb92a563-c01f-4bbb-96a6-25bae70aa06f.png)

PR #519, opened 2022-08-05
fix: make the component work in swarm-extension

vojtechsimetka:
Somewhat addresses #518 . I am sure better solution could be designed, but with this the Swarm Extension works.


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgza4wvvxnd2niaif3yl6dung3a3ifwwhz5rrg3dsql6gc4oynunal2q.bzz.link
**Swarm Hash:** `e5ab5bb47a6a0082ef0bf0e8d36c1b416d63e7b189b639417e30b8ec368d02f5`
**Commit Hash:** fe87ccde4ed971f7edb66f4e4e48d70a8a21a65f
**Commit Message:** fix: make the component work in swarm-extension

src/config.ts
```diff
@@ -1,3 +1,9 @@
+// FIXME: in the Swarm Extension process is not defined and would throw an error
+// see https://github.com/ethersphere/bee-dashboard/issues/518
+export function getEnvVariable(variableName: string): string | undefined {
+  if (typeof process === 'object') return process.env[variableName]
```
AuHau, line 4: Did you test this for the normal Dashboard? I remember that the test `typeof process` did not work with the CRA mechanism of the env. replacement. Maybe you have to extend it to `process.env`...

vojtechsimetka: You are right. However, the test `process && process.env` complaints in the Swarm Extension so I wonder if removing the env variables here and injecting them instead would not be easier.

Cafe137:
Swarm Extension has been released with v0.20.2

PR #530, opened 2022-09-02
feat: error reporting callback

### LINKED ISSUE #524: Add way to report errors "upstream" when used as component
With the component refactor in Desktop, we track errors using Sentry on top-level. This unfortunately might not work as good, because we might not catch the errors that were originally caught by the Sentry's ErrorBound that was removed in #520. We could create a callback that could be passed as the Dashboard's App component's prop that would get called when this errors are caught here: https://github.com/ethersphere/bee-dashboard/blob/0260df61de0619202a819b79820cfbef6e3757ae/src/components/ErrorBoundary.tsx

AuHau:
Closes #524 

Using `refactor/removal-env-variables` branch as a base only for better readability.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,9 +22,17 @@ interface Props {
   lockedApiSettings?: boolean
   isDesktop?: boolean
   desktopUrl?: string
+  errorReporting?: (err: Error) => void
 }
 
-const App = ({ beeApiUrl, beeDebugApiUrl, lockedApiSettings, isDesktop, desktopUrl }: Props): ReactElement => {
+const App = ({
+  beeApiUrl,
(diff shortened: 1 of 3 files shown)


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzaocdvl27qgqwh64cta3v6szase3e7v2dewr6owx3mti6jo2h4it4q.bzz.link
**Swarm Hash:** `708755ebf0342c7f705306ebe9641226c9fae864b47ceb5f6c9a3c9768fc44f9`
**Commit Hash:** 8a322573c4e64795267b5c1e171aa310ea286c5e
**Commit Message:** feat: error reporting callback

PR #588, opened 2022-11-15
feat: add prerequisite checks before swap

Cafe137:


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzavdazz3sszpxytc7u73octjz22mb6srti5owqn4vq6w45j5fzcqia.bzz.link
**Swarm Hash:** `a8c19cee52cbef898bf4fedc29a73ad303e94668ebad06f2b0f5b9d4f4b91410`
**Commit Hash:** 1d32c6618756a181cca3851ef4f7c4ecf275c7fe
**Commit Message:** feat: add prerequisite checks before swap

src/pages/top-up/Swap.tsx
```diff
@@ -127,6 +134,27 @@ export function Swap({ header }: Props): ReactElement {
     }
   }
 
+  async function performSwapWithChecks(daiToSwap: DaiToken) {
```
AuHau, line 137: Sooo I am not really content with how the error handling is done. I would suggest removing all the `enqueueSnackbar` cases here and leaving it to the top-level `try/catch` in the `onSwap()` function to do the error presentation.

I would also suggest creating a custom Error something like `SwapError` that would have a property  `snackMessage` or similar, where the custom error messages would be placed.

Then I would suggest a helper function called something like `wrapErrorWithMessage<T>(promise: Promise<T>, errorMessage: string): Promise<T>` that would catch an error and would wrap it with the `SwapError` with your snack message and rethrow. In this way, you avoid the `.catch()` handling.

src/pages/top-up/Swap.tsx
```diff
@@ -127,6 +134,27 @@ export function Swap({ header }: Props): ReactElement {
...
+    const desktopConfiguration = await getDesktopConfiguration(desktopUrl).catch(error => {
+      enqueueSnackbar('Unable to reach Desktop API. Is Swarm Desktop running?', { variant: 'error' })
+      throw error
+    })
+
+    if (!desktopConfiguration['swap-endpoint']) {
+      const message = 'Swap endpoint not configured in Swarm Desktop'
+      enqueueSnackbar(message, { variant: 'error' })
+      throw Error(message)
```
AuHau, line 146: Here you would simply throw `SwapError`

AuHau (commented):
Good idea for consolidating the desktop configuration into `desktop.ts`! That said, I would prefer better-structured error handling in the `onSwap()`. My suggestion is below.

src/pages/top-up/Swap.tsx
Cafe137: OK I generally agree, what I want to achieve here is to have the full raw error logged to console, and a nicer and clear error message on the screen. I'll implement both with your approach.

AuHau: Nice 👍

src/pages/top-up/Swap.tsx
```diff
@@ -135,16 +162,23 @@ export function Swap({ header }: Props): ReactElement {
...
-      console.error(error) // eslint-disable-line
-      enqueueSnackbar(`Failed to swap: ${error}`, { variant: 'error' })
+      if (isSwapError(error)) {
+        enqueueSnackbar(error.snackbarMessage, { variant: 'error' })
+
+        if (error.originalError) {
+          console.error(error.originalError) // eslint-disable-line
+        }
+      } else {
+        console.error(error) // eslint-disable-line
```
AuHau, line 180: I would still show some snackbar even if it is not `SwapError`. It could be either a generic one something like "Ups, something went wrong", or you could add the `error.message`... 
Generally, it is a bad idea to have a case where you swallow an error without letting the user know (and IMHO `console.error` is not "letting the user know") especially if he expects some action...

AuHau (changes requested):
Generally big improvement! But still one more issue ;-)

src/pages/top-up/Swap.tsx
Cafe137: Yup, totally missed this case, thanks

AuHau (approved):
LGTM 👍 Thanks

PR #593, opened 2022-11-24
fix: handle auth and server error during swap

Cafe137:


bee-worker:
## 🐝 PR preview in Swarm

**Preview URL:** https://bah5acgzatunrqzhzazvhfzobmzmqumalf5ticn77j4jcwei2uwirrd6zf3ra.bzz.link
**Swarm Hash:** `9d1b1864f9066a72e5c166590a300b2f668137ff4f122b111aa591188fd92ee2`
**Commit Hash:** b5c8abb6d54475aca475f707db0a1197d65e959c
**Commit Message:** fix: change execution order for light node upgrade

src/pages/top-up/Swap.tsx
```diff
@@ -126,31 +126,50 @@ export function Swap({ header }: Props): ReactElement {
...
   async function performSwapWithChecks(daiToSwap: DaiToken) {
-    const desktopConfiguration = await wrapWithSwapError(
+    if (!localStorage.getItem('apiKey')) {
+      throw new SwapError('API key is not set, reopen dashboard through the desktop app')
+    }
+
+    let desktopConfiguration = await wrapWithSwapError(
       getDesktopConfiguration(desktopUrl),
-      'Unable to reach Desktop API. Is Swarm Desktop running?',
+      'Unable to reach Desktop API - Swarm Desktop may not be running or API key is wrong',
```
AuHau, line 155: We should be able to distinguish if the API key is wrong or Swarm Desktop server is not running, no?

src/utils/net.ts
```diff
@@ -27,6 +27,12 @@ export async function sendRequest(
     method,
     headers,
     data,
+  }).catch(error => {
+    if (error?.response?.data) {
+      throw Error(JSON.stringify(error.response.data))
```
AuHau, line 32: I would provide a bit nicer Error 😅 At least format it like suggested. Maybe include path of the request as well?
```suggestion
      throw Error(`Request failed! Response: ${JSON.stringify(error.response.data)}`)
```

AuHau (approved):
Generally LGTM, some suggestions which I will leave up to you.

Issue #606, opened 2023-01-30
Entering * into amount when buying a postage stamp returns error


crtahlin:
Trying to enter an "*" into Amount field, returns an error. See screenshots.

![image](https://user-images.githubusercontent.com/1554520/215465247-ac953a0e-5598-4e67-8686-c3e2e0e693d8.png)

![image](https://user-images.githubusercontent.com/1554520/215465300-4d04fa2b-511b-4a92-97b1-be8225f32838.png)

Expected: just allow numbers to be entered?

Cafe137:
This issue comes from PostageStampCreation.

Issue #630, opened 2023-10-24
Stamp Labels are missing in the "Add postage stamp" page of update feeds


rampall:
![image](https://github.com/ethersphere/bee-dashboard/assets/520570/89587252-5622-449d-8adc-925d685cfe8c)
![image](https://github.com/ethersphere/bee-dashboard/assets/520570/a1f3fb66-03cb-4545-9459-69c6184a97ef)

Cafe137:
This is in PostageStampSelector.

If a label is set on a stamp, it should show the label, and then the short hash in parantheses.

PR #634, opened 2023-11-03
fix: add missing stamp labels and fix inputs

zol1981:
Fixed issues:
#606 Entering * into amount
#630 Stamp Labels


zol1981:
found another bug, can't use dots (bignumber.js isInteger function only works when typing digits after dots, but bigInt fails when number contains a digit, like num = 12. num.isInteger(num) true but bigInt(num) fails)

PR #647, opened 2023-12-13
feat: show syncing info

ferencsarai:


ferencsarai:
@ethersphere/solar-punk

PR #648, opened 2023-12-15
allow console.log

Baleba:
First almost working.
Can create feeds with topic other than website.
Can store json data in feed posts.
Can read previous posts 

Can choose an existing ID to work on feeds.


Baleba (commented):
Temporary of course

Issue #432, opened 2022-06-22
Node is not connected shown in case of a warning


agazso:
`Node is not connected` shown on the info page, even if there is just a warning on the status page.

Example:
<img width="1353" alt="Screenshot 2022-06-22 at 17 56 55" src="https://user-images.githubusercontent.com/230163/175077932-c57d7144-34a2-4608-bc5f-360ab7158a91.png">
<img width="1353" alt="Screenshot 2022-06-22 at 17 57 36" src="https://user-images.githubusercontent.com/230163/175077948-f566888f-bacd-497d-b991-c22fa2ac5652.png">

agazso:
This is still an issue in the current version, using desktop 0.15.0-pre.14

PR #700, opened 2025-07-21
fix: move status to state

Kexort:
Both the SideBarStatus and NodeInfoCard will now reflect the same status state since they're both using the same stateful status from the context.


Kexort:
Resolves #432

